PR #17, opened 2026-06-21
perf(upload): single 60s flusher as the sole upload trigger

abdulsaheel:
## What

Collapses the six `upload()` triggers in `AppState` (one 15s periodic flusher + five event-driven calls: foreground-reclaim, post-drain, reconnect, forceResync) into **one always-running flusher**. It kicks once on start, then uploads everything pending every **60s**.

## Why

All record sources — live `0x28`/`R10`/`0x33` and historical drain — already store locally to `raw_records` (hex PK, `INSERT OR IGNORE`, retain-until-200). The only thing that varied was *when we POST*. Six triggers meant six things to keep in sync. This makes the 60s flusher the **single** thing that POSTs, so upload behaviour is uniform everywhere — one path: if it breaks, it all breaks; if it works, it works everywhere.

## Cost impact

The 15s timer turned the steady-state ~1 Hz trickle into a POST every 15s (~**5,760/day**). At 60s that's ~**1,440/day** — **~4× fewer** R2 Class A puts + D1 row writes + Workers requests (R2 puts and D1 writes are 1:1 with POSTs). Roughly **\$1.15 → \$0.30 per user/mo at 1,000 users**. The `_kFlushInterval` constant can be raised further (e.g. 5 min → ~290/day, ~\$0.08/user) at no fidelity cost if more latency is acceptable.

## Safety / invariants preserved

- **Delete-by-hex on confirmed 200** (retain-until-200) is unchanged — the no-data-loss invariant. A non-200/429 leaves rows queued for the next tick.
- **Raw-first** `_flushDrain()` (local write before ACK) untouched.
- **iOS headless-recovery** (`runHeadlessSync`) keeps its direct `uploadPending()` — it runs in a separate background isolate where `AppState`/the timer don't exist, but uses the *same* primitive + same delete-by-hex semantics.

## Behaviour change (accepted)

Freshly-drained data now waits ≤60s for the next tick instead of uploading instantly. Invisible to cloud-derived metrics (HRV/recovery/etc. are computed at the wake-close, not on arrival).

## Edge cases covered

| Case | Behaviour |
|---|---|
| Network down / timeout / 429 | non-200 → rows retained → retried next tick |
| App killed mid-upload | only confirmed-200 hexes deleted; rest upload next session |
| Backlog after long disconnect | one tick drains it all in 300-record chunks |
| Live + drain overlap (same second) | hex PK dedupes locally; backend dedupes by (user,device,counter) |

## Testing

- `flutter analyze` clean.
- ✅ **Device-verified** (real band → v2 backend): kick-on-start (`Uploaded 159/159 records`), steady-state 60s tick with multi-chunk loop (`Uploaded 743/743 records` = 300+300+143), events path (`Uploaded 20 events`), and delete-by-hex/retain-until-200 (pending drains to 0, nothing stuck). Connect → drain → live → keep-alive lifecycle unaffected.

### DIFF
--- a/lib/state/app_state.dart
+++ b/lib/state/app_state.dart
@@ -79,23 +79,19 @@ class AppState extends ChangeNotifier {
   /// notification and the live drain + flush continue.
   bool _background = false;
 
-  /// One session-long flusher. It uploads pending raw records on a steady cadence and,
-  /// because the uploader RETAINS rows on any non-200 (a transient rate-limit 429
-  /// included), each tick also retries whatever the last tick couldn't send. No
-  /// per-record flushing and no backoff/cooldown — a plain ~15s cadence sits well under
-  /// the backend rate limit (burst 30, refill 0.5/s) on its own. On iOS the timer simply
-  /// fires on the next BLE-notification resume when the app was suspended.
   // THE single upload path. Every record source (live 0x28/R10/IMU, historical drain)
   // only ever STORES locally (raw_records, hex PK, retain-until-200). This one timer is
   // the ONLY thing that POSTs: it kicks once on start, then uploads everything pending
-  // every _kFlushInterval, deleting each chunk only on its confirmed 200. Uniform
+  // every _kFlushInterval, deleting each chunk only on its confirmed 200 (so a non-200/
+  // 429 just leaves rows queued for the next tick — no backoff needed; the 60s cadence
+  // sits well under the backend rate limit of burst 30 / refill 0.5/s). Uniform
   // behaviour everywhere — no per-event upload() calls to drift out of sync.
   //
   // The 60s cadence collapses the steady-state 1 Hz trickle from ~1 POST/15s (~5,760/
   // day) to ~1 POST/min (~1,440/day) — ~4× fewer R2 puts + D1 writes + Workers requests,
   // with zero data loss (cloud metrics land at the wake-close, so ≤1 min upload latency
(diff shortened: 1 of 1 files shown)


abdulsaheel:
Device-verified on hardware (real band → v2 backend), happy path end-to-end:

- `Uploaded 159/159 records.` — kick-on-start uploaded the drained backlog immediately (the path that replaced the removed explicit post-drain `upload()`).
- `Uploaded 743/743 records.` — steady-state 60s flusher tick, multi-chunk loop (300+300+143), all accepted.
- `Uploaded 20 events.` — events path.
- Pending count drains to 0; delete-by-hex / retain-until-200 holding.

Core risk (uploads firing via the single flusher rather than the removed per-event calls) is retired. Ready to merge.

abdulsaheel:
thanks @Copilot — all three are fair, they're doc drift not logic (behaviour's correct + device-verified). fixed in e63a5d8:

- killed the stale `~15s cadence` block that got left stacked above the new 60s one
- `_startFlusher`: clarified the kick only fires on a cold start — on reconnect/foreground the timer's still alive so it's a no-op and the next tick (≤60s) carries it, not "without waiting a full interval"
- corrected "stopped only by endSession" → it's stopped on session teardown (logout/signOut/unpair/endSession)
- foreground-reclaim comment no longer says "and flush" since the always-running flusher keeps it current

PR #25, opened 2026-06-27
improved sync, storage tables and compute

localhoop:


abdulsaheel:
yeah copilot's got a point on the perf but honestly thats the small part. the bigger thing for me is we now have a second copy of the hrv math living in here - _noopSessionAvgHrv / _noopCleanRr / _noopRejectEctopic / _rmssdRaw are basically redoing what openstrap_analytics already does (correctRr, rmssd, the ectopic rejection). two versions of the same physiology will drift sooner or later and the package one is the tested one, so id rather these call into ana.* than keep their own copy.

on the actual loop: yep it rescans the whole rr array for every 5min window, and even scans beats outside start/end each pass, so O(windows x beats). for a normal night its only a couple ms so not urgent, but rrTsMs is already time sorted so one pass bucketing by (tsSec-start)~/300 gets it to O(n). easy fix.

not blocking from my side, happy for it to be a followup. just wanted to flag the duplication before it spreads.

localhoop:
I have resolved the AI comments and moved the HRV to analytics: https://github.com/OpenStrap/analytics/pull/10

PR #65, opened 2026-07-11
feat: on-demand wear-location check (GET_BODY_LOCATION_AND_STATUS)

abdulsaheel:
## Summary
WHOOP garments beyond the wrist strap (bicep band, apparel pocket) mean the band firmware itself classifies which body location it's currently docked in. The protocol layer already had the command builder (`cmdGetBodyLocationAndStatus`, opcode `0x54`, **non-dangerous**) and response parser (`BodyLocationStatusResponse` / `GarmentDeviceLocation` enum) built from APK ground-truth — edge never actually sent the command or surfaced the result anywhere. This wires it end-to-end.

## Changes
- `ble_engine.dart`: `BleEngine.getBodyLocationAndStatus()` sends the command; `_absorbState` handles the `body_location_status` response field into new `DeviceState` fields.
- `models.dart`: `DeviceState.bodyLocationRaw/Confidence/Status/CheckedAt` — raw ints only. `confidence`/`status` are unconfirmed-scale firmware bytes (APK ground-truth only surfaced the command + field layout, not their exact semantics) — never re-interpreted past what's actually confirmed.
- `app_state.dart`: `AppState.checkBodyLocation()` — **user-triggered only**, deliberately NOT sent automatically on connect (keeps the sync-critical handshake path untouched) — plus a `bodyLocationLabel` getter resolving the raw byte into a friendly string, defaulting honestly to "Not checked" / "Not conclusive" rather than assuming "Wrist" just because that's what most users own.
- `profile_screen.dart`: new "Wear location" fact + "Check" action on the strap details tile, right next to the existing Battery/Wrist/Serial facts — mirrors the existing "Sync now" pill button pattern exactly.

## Test plan
- [x] `flutter test --concurrency=1` — 438/445, same 7 pre-existing unrelated failures as clean main
- [x] Updated `DeviceTile` tests: fact renders, "Check" tap wiring, action hidden while disconnected

### DIFF
--- a/lib/ble/ble_engine.dart
+++ b/lib/ble/ble_engine.dart
@@ -1817,11 +1817,20 @@ class BleEngine {
     }
     if (f.containsKey('body_location_status')) {
       final b = f['body_location_status'] as BodyLocationStatusResponse;
+      state.bodyLocationRevision = b.revision;
       state.bodyLocationRaw = b.locationRaw;
       state.bodyLocationConfidence = b.confidence;
       state.bodyLocationStatus = b.status;
       state.bodyLocationCheckedAt =
           DateTime.now().millisecondsSinceEpoch ~/ 1000;
+      // Logged unconditionally (not just on an unmapped value) — this command
+      // is new and user-triggered rarely, so a real-world raw-byte trail is
+      // worth keeping regardless of whether GarmentDeviceLocation resolves it.
+      _log(
+        'GET_BODY_LOCATION_AND_STATUS: rev=${b.revision} '
+        'locationRaw=${b.locationRaw} (${b.location?.name ?? "UNMAPPED"}) '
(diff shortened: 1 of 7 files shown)


abdulsaheel:
Abandoning this feature. Real device data (body_location_checks captures across two sessions) shows the response's \`revision\` byte behaving like a rolling activity/session counter rather than a stable protocol constant (162→167→170→172 within one burst, then a LATER session independently landing on the same 170→172 tail) — meaning our assumed byte offsets for this response may be wrong, not just that the band is honestly reporting low confidence. Without raw-hex capture to actually verify the offsets, we can't tell whether "unknown location" is a genuine band-side report or a decode bug, and it's not worth carrying the uncertainty/maintenance burden for a feature this minor. Not shipping it.

PR #68, opened 2026-07-11
fix: GPS route noise floor + map-mode layout harmony; add gallery preview

abdulsaheel:
## Summary
Fixes three related bugs reported against GPS workouts (run/ride/walk):
1. Map defaults zoomed in too far to even render — required a manual zoom-out.
2. Tiny indoor movements (shifting position in bed) got counted as real distance/route.
3. The live map's placement/layering looked badly composed.

## Root cause (1+2, same cause)
`RouteTracker._onSample` filtered on accuracy and implausible jumps, but had **no minimum-movement noise floor** — ordinary GPS multipath jitter while stationary (a few metres, well within a "good" reported accuracy) passed both gates and got added to distance + the route polyline unconditionally. That jittery cluster then degenerated the route's `LatLngBounds` to near-zero size, which made the map's fit-to-bounds zoom in to near-max trying to "fit" it.

**Fix**: `RouteTracker` gains `minMovementM` (default 4m — an engineering choice like the existing `maxJumpM`/`maxSpeedMps`, not a cited threshold). A fix within the floor of the last **accepted** point is dropped without advancing the anchor, so isolated jitter never nets a displacement and is dropped for good — but genuine slow movement still accumulates across drops and gets captured (coarsely) once it clears the floor.

## Root cause (3)
The live map wasn't a purpose-built GPS layout — it was a small boxed overlay floating on top of the ember/HR-reactive "core" screen (separate zone-tinted background, ember particle field, HR circle, and zone ladder all still rendering underneath). Two competing UI systems sharing the same Stack is what actually read as badly layered.

**Fix**: extracted `GpsLiveMapView` — a pure/previewable widget with ONE unified Strava-style bottom stat bar (distance, duration, pace, BPM zone-coloured). For a GPS workout the map now **is** the screen (full-bleed), not an inset box; the ember core is suppressed in map mode instead of competing underneath it. Non-GPS workouts are untouched.

## Finish screen
For a GPS workout, the recorded route is now the **first** thing shown after the header (reusing `RouteCard`, the same widget the workout detail screen already uses) — not a small 140px thumbnail buried at the end among the strain/zone/PR cards. Strava-style "map is the hero."

## Native maps — discussed, staying with flutter_map + CARTO
Not switching to Apple/Google native maps: would lose the single-`ColorFilter` branding trick and add two native integration surfaces + a Google Maps API key/billing dependency, for a real but smaller quality/perf win. Revisit separately if it still bothers us later.

## Design Gallery
New "Workout preview" section — a deterministic fake ~3.2km/20min run route (`fake_route_fixture.dart`, distance/moving-time/splits all **derived** from the generated points via the real `route_math.dart` functions, never hand-typed) so both the live GPS map layout and the finish screen's route-hero layout can be reviewed without a live device, GPS fix, or BLE connection. `WorkoutFinishScreen` gained an optional `previewRoute`/`previewMaxHr` to support this.

## Test plan
- [x] `flutter test --concurrency=1` — 448/450 real tests (the 2 "failures" are untracked scratch/debug files, not real code)
- [x] `dart analyze` — clean on every touched file
- [x] New `route_tracker_test.dart` coverage: drops stationary jitter (no distance/vertex), genuine slow movement still counted, a route dominated by stationary jitter keeps a real tight bounding box
- [x] New `fake_route_fixture_test.dart`: fixture produces a well-formed, plausible, deterministic route
- [x] Existing `workout_sleep_redesign_test.dart` (WorkoutFinishScreen snapshot-only render) still green

### DIFF
--- a/lib/gps/route_tracker.dart
+++ b/lib/gps/route_tracker.dart
@@ -38,6 +38,26 @@ class RouteTracker {
   /// spike is still rejected.
   final double maxJumpM;
 
+  /// GPS noise floor (metres): a fix within this distance of the last
+  /// ACCEPTED point is dropped — not added to distance or the route polyline.
+  /// Consumer GPS commonly drifts a few metres between fixes from multipath
+  /// alone, even while genuinely stationary (indoors, or before a workout's
+  /// real movement starts) and even with a "good" reported accuracy well
+  /// under [maxAccuracyM]. Without this floor, that drift silently
+  /// accumulates into real-looking phantom distance, AND — separately —
+  /// pollutes the route's bounding box with a tight jittery cluster, which
+  /// made the map's fit-to-bounds zoom in to near-max trying to "fit" it.
+  ///
+  /// Deliberately does NOT advance the anchor when a fix is dropped: the next
+  /// fix is compared against the SAME last-accepted point, so genuine slow
+  /// movement still accumulates and gets captured (just in coarser steps)
+  /// the moment its cumulative distance from the anchor clears the floor —
+  /// only fixes that never net a real displacement (isolated back-and-forth
+  /// jitter) are dropped for good. An engineering choice tuned to typical
+  /// phone-GPS drift, not a cited scientific threshold — same convention as
+  /// [maxJumpM]/[maxSpeedMps].
+  final double minMovementM;
+
   /// Fastest plausible sustained speed (m/s) for the jump allowance.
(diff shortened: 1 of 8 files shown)


abdulsaheel:
Closing — redesigning per clarified direction before this goes anywhere near main. Removing the branch/commits from the public repo per request.

PR #71, opened 2026-07-11
fix: Siri query staleness + add StartBreathingIntent action

abdulsaheel:
## Summary
Two things: (1) root-cause fix for "Siri fails a lot in early testing", (2) the first real Siri **action** intent (not just query), tying into the just-shipped cardiac-coherence feature.

## 1. Root cause of Siri staleness/failure
`OpenStrapIntents.swift`'s `RecoveryIntent`/`StrainIntent`/`SleepIntent` read `readiness`/`strain`/`hrv`/`sleep_min`/`rhr` from the App Group. **The only place that ever wrote those keys was `today_screen.dart`'s `fetch()`** — i.e. only when the user had recently opened the Today screen in the foreground app. "Ask Siri instead of opening the app" is the entire point of a Siri Shortcut, so any query made without a recent Today-screen visit answered from stale data (or hit the generic "I don't have today's numbers yet" fallback) — reading exactly like "Siri is broken."

**Fix**: `ios_bg_task.dart` now pushes a fresh Today snapshot after every headless derive pass (both the heavy BGProcessingTask and light BGAppRefreshTask profiles), via a standalone `LocalRepositoryImpl` (no Provider/BuildContext needed, same headless-safe construction this file already uses elsewhere). Best-effort, never throws into the caller.

## 2. New action intent: StartBreathingIntent
Unlike the query intents, starting a breathing session needs the live Flutter engine + BLE stack (real RR from the band), so `openAppWhenRun = true`. It writes a pending route into the App Group (mirrors the existing `end_session` Live-Activity-intent pattern), consumed on the Dart side via a new `WidgetService.consumePendingRoute()`, fed through the **same** tap-route deep-link pipeline notifications already use (`tap_router.dart`'s new `kRouteBreathing` → `CalmBreathingScreen(autoStart: true)`). Checked at cold launch (`AppState`'s constructor) **and** on every foreground resume (`app.dart`), since `openAppWhenRun` may just foreground an already-running process rather than a fresh launch.

New phrases: "Start breathing in OpenStrap", "OpenStrap breathe", "Start a breathing session in OpenStrap", "Breathe with OpenStrap".

## Test plan
- [x] `dart analyze` on all touched files — clean (2 pre-existing unrelated warnings)
- [x] `flutter test --concurrency=1` — 454/456 real passes (450 + 4 new `tap_router_test.dart` cases), same 2 pre-existing scratch-file failures as clean main (not real code)
- [x] `flutter build ios --debug --no-codesign` — succeeds, new Swift intent compiles cleanly
- [ ] **Not verified**: actual on-device Siri recognition latency/reliability for the new phrase — this is real OS-side behavior outside static verification, needs a real-device test

### DIFF
--- a/ios/OpenStrapIntents.swift
+++ b/ios/OpenStrapIntents.swift
@@ -80,6 +80,29 @@ struct SleepIntent: AppIntent {
   }
 }
 
+// MARK: - Action intents (these actually DO something, not just answer)
+
+/// "Start breathing" — unlike the query intents above, this needs the live
+/// Flutter engine + BLE stack (a guided session reads live RR from the band),
+/// so it must open the app rather than answer standalone. Writes the target
+/// route into the App Group; the Dart side picks it up via
+/// WidgetService.consumePendingRoute() on launch AND on every foreground
+/// resume (see AppState.checkPendingSiriRoute — openAppWhenRun doesn't
+/// guarantee a fresh launch, it may just foreground an already-running
+/// process, so both call sites matter).
+@available(iOS 16.0, *)
+struct StartBreathingIntent: AppIntent {
+  static var title: LocalizedStringResource = "Start Breathing Session"
+  static var description = IntentDescription(
+    "Start a guided resonance-breathing session in OpenStrap.")
+  static var openAppWhenRun = true
+
+  func perform() async throws -> some IntentResult & ProvidesDialog {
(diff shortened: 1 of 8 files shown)


abdulsaheel:
Superseded by #72 — turns out that branch already includes all of this PR's commits in its history (never branched fresh between the two features), so #72 is the single combined PR covering both the Siri fix and the Live Activity feature. Closing this one to avoid confusion.

PR #75, opened 2026-07-12
update screenshots, they were 3 weeks stale

abdulsaheel:
replaced with fresh ones off a real device today. old set still had a
Notifications screen (removed a while back), an AI Coach screenshot, and
activity/strain/stats as separate things even though Body covers all of
that now.

new set: today, sleep (hypnogram actually showing real stages), heart,
stress (new screen), breathing (the resonance-breathing screen with the
real coherence readout - shows the honest "calibrating" state, not a fake
number), body, steps, workouts, records, recap, profile, plus widget /
battery-widget / live-activity for ios.

a couple of things worth calling out on the source photos, in case anyone
wonders why some are missing:
- cropped the live-activity screenshot down to just the widget - the
  original had a personal photo in the background
- skipped a gps live-workout screenshot entirely - it was a quick 12-second
  test with basically zero real data, and the map tile rendered a real
  street/neighborhood name, not worth keeping either way

resized everything to the same ~340px width the old set used and
compressed - file sizes are comparable or smaller across the board despite
being real device screenshots now instead of whatever the old ones were.

this is a docs/assets-only PR, branched separately from the bug-fix PR
(#74) so they don't collide. not merging myself, wanted a look first.


abdulsaheel:
Merging the screenshot update into #74 instead so both land in one PR.

PR #87, opened 2026-07-17
Fix first-run blockers for external contributors on current Flutter and iOS

alexforia:
- theme.dart: import CupertinoPageTransitionsBuilder from cupertino (no longer re-exported by material on Flutter 3.44+)
- firebase_options.dart: use a format-valid dummy API key; FirebaseInstallations validates the key format natively and the resulting NSException cannot be caught from Dart, crashing the app at launch on device
- Watch app entitlements: use $(APP_GROUP_IDENTIFIER) instead of the hardcoded group.wtf.openstrap, which breaks signing for any external team
- OpenStrapBatteryWidget: read the app group from AppGroup.identifier instead of a hardcoded constant
- pairing_screen: offer the AccessorySetupKit picker before checking bluetoothReady(); before any ASK-provisioned accessory exists, iOS reports the adapter as unauthorized, so the old order deadlocked pairing behind a false 'Bluetooth is off' screen
- app_state: bluetoothReady() now waits for a determinate adapter state instead of grabbing the initial 'unknown'

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bluetooth readiness detection during device pairing.
  * Pairing now correctly prioritizes supported iOS 18 accessory selection flows.
  * Prevented pairing from being blocked by Bluetooth’s initial startup state.
  * Improved reliability of shared data between the app and watch/widget.
  * Updated Firebase configuration handling to avoid startup issues with platform validation.

### DIFF
--- a/ios/OpenStrapWatch Watch App/OpenStrapWatch Watch App.entitlements
+++ b/ios/OpenStrapWatch Watch App/OpenStrapWatch Watch App.entitlements
@@ -4,7 +4,7 @@
 <dict>
 	<key>com.apple.security.application-groups</key>
 	<array>
-		<string>group.wtf.openstrap</string>
+		<string>$(APP_GROUP_IDENTIFIER)</string>
 	</array>
 </dict>
 </plist>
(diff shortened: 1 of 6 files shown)


abdulsaheel:
thanks, great addition !

PR #95, opened 2026-07-18
feat(notify): buzz the strap for incoming phone calls

### LINKED ISSUE #92: Buzz Notifications - Missing phone call app
Thanks for making this - it really is excellent!

I've got buzz notifications working for my messaging apps on Android, but there's no 'phone' option when selecting the apps. 

Is there any way of enabling buzz notifications for incoming phone calls on Android?

dannymcc:
Closes #92.

## Why the relay can't do this

Two separate blockers mean incoming calls can never ride the existing notification relay:

1. The app picker uses `InstalledApps.getInstalledApps(excludeSystemApps: true)`, and the dialer (Google Phone, Samsung Phone, AOSP Dialer) is a system app on essentially every device — so there's no 'Phone' entry to select (exactly what #92 reports).
2. Even if it were listed, `NotificationRelay._onNotification` skips ongoing notifications (`if (e.hasRemoved || e.onGoing) return;`) — correctly, to avoid media players and foreground services — and Android posts incoming-call notifications as *ongoing* full-screen-intent notifications, so ringing is precisely the kind of notification that filter swallows.

## What this does instead

Calls get their own telephony-based path:

- **`CallStateBridge.kt`** — a small native bridge streaming `ringing` / `offhook` / `idle` from `TelephonyManager` over an EventChannel (`TelephonyCallback` on API 31+, `PhoneStateListener` below). Registered on the long-lived engine alongside `NativeChannels`, so it keeps working while backgrounded/headless. Permission request borrows the Activity `CompanionBridge` already tracks; the result is forwarded from `MainActivity.onRequestPermissionsResult`.
- **`CallBuzzer`** (Dart) — same persisted-ChangeNotifier idiom as `NotificationRelay`. While RINGING it buzzes the band immediately and then every 4 s — a cadence on the wrist reads unmistakably as 'your phone', where a single pulse is easy to miss — and stops the instant the call is answered, declined, or rings out. A cap of 8 buzzes (~one standard 30 s ring) guards against OEMs that never emit IDLE after a missed call.
- **UI** — a 'Buzz on incoming calls' card on the Band-notifications screen, independent of the app-relay master toggle (it doesn't need notification access), with inline permission grant and a truthful profile-card subtitle ('On · calls only', '… · calls').

## Permissions

Adds `READ_PHONE_STATE` (runtime, requested only when the toggle is switched on). Only the call *state* is read — numbers never are (no `READ_CALL_LOG`), and the manifest comment says so. iOS is untouched: no ringer-observation API exists there, so the feature is inert and hidden, matching the relay's pattern.

## Testing

- `flutter analyze` clean for the touched files; `flutter test` — the new `call_buzzer_test.dart` covers the cadence (immediate buzz, repeat interval, stop on offhook/idle, stuck-RINGING cap, duplicate-event and re-ring behaviour) under `fakeAsync`. The two pre-existing `derivation_pipeline_test.dart` failures on main are unrelated (git-dep drift) and unaffected.
- Not yet verified on a physical device — happy to iterate if anything misbehaves on OEM dialers.

## Summary by CodeRabbit

* **New Features**
  * Added an Android-only incoming-call buzz alert that taps the strap when calls are ringing.
  * Introduced runtime phone-permission handling to enable call buzzing and prompt for access when needed.
  * Added a call-buzz settings card and updated the notification relay section to reflect combined relay/call states.
  * Implemented native call-state monitoring and automatic start/stop of buzzing based on call transitions.
* **Tests**
  * Added automated tests covering ringing cadence, stopping on answered calls, repeat caps, and connection-aware behavior.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -22,6 +22,9 @@
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
     <!-- Band-gesture "ring my phone" haptic. Normal permission, no runtime prompt. -->
     <uses-permission android:name="android.permission.VIBRATE" />
+    <!-- Incoming-call strap buzz: observe the RINGING/OFFHOOK/IDLE call state only.
+         Runtime permission; numbers are never read (no READ_CALL_LOG). -->
+    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
     <!-- Reboot survival: restart EdgeTrackingService + headless auto-connect after device reboot. -->
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <!-- CompanionDeviceManager: association makes us a "companion app" (exempt from the
--- a/android/app/src/main/kotlin/wtf/openstrap/openstrap_edge/CallStateBridge.kt
+++ b/android/app/src/main/kotlin/wtf/openstrap/openstrap_edge/CallStateBridge.kt
@@ -1,0 +1,162 @@
+package wtf.openstrap.openstrap_edge
+
+import android.Manifest
+import android.content.Context
+import android.content.pm.PackageManager
+import android.os.Build
+import android.os.Handler
+import android.os.Looper
+import android.telephony.PhoneStateListener
+import android.telephony.TelephonyCallback
+import android.telephony.TelephonyManager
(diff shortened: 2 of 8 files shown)


dannymcc:
I actually think the Tasker integration makes this unnecessary now.

PR #96, opened 2026-07-19
feat(widget): Android home-screen widgets (metrics + band battery)

dannymcc:
The iOS side has a lovely set of WidgetKit widgets; Android had none — even though `WidgetService` already writes every snapshot key via `home_widget` and broadcasts to an (until now nonexistent) `OpenStrapWidgetProvider`. This adds the Android siblings as classic RemoteViews providers, **no new dependencies**.

Rendered preview of the Android widgets

*Rendered preview (not a device screenshot) — generated from the same palette, geometry and ring maths as the code, using the iOS placeholder values.*

## What's included

- **`OpenStrapWidgetProvider`** — Ready · Strain · Sleep · HRV rings. Two faces like the iOS families: a 2×2 grid when small, a single four-ring row when wide. The iOS medium stacks a readiness headline over the rings, but an Android 4×2 cell only gets ~110 dp of height (vs systemMedium's 155 pt), so the wide face puts the four rings in one row with the readiness value carrying the readiness colour instead. Layout switching uses the RemoteViews size map on Android 12+, min-width fallback below.
- **`OpenStrapBatteryWidgetProvider`** — band battery ring from the last BLE snapshot (batt_pct/batt_charging/batt_name), charging bolt in the label, and the same >24 h staleness mute as the iOS widget. Never fetches — battery is a live BLE value only the app knows.
- **`StrapWidgets`** — shared palette, snapshot readers, and a Canvas arc renderer (RemoteViews can't draw arcs, so rings are pre-rendered bitmaps at the device density). All colour/threshold/format rules mirror the Swift widgets 1:1 — including Ember-on-Paper/Char theming via the `theme_dark` key the app already writes, HRV green-at-baseline shading, and the `hm()` sleep format.
- **Dart** — `pushBattery`/`setThemeDark` now also target the battery provider; the metrics provider name was already wired upstream.

## Notes for review

- **home_widget storage gotcha:** on Android the plugin stores Dart doubles as **raw long bits** (`putLong(doubleToRawLongBits(...))`), so `strain` must be read back with `Double.fromBits` — `StrapWidgets.readDouble` encodes this (and coerces defensively through `all[key]` so a type drift can never crash a widget).
- No self-refresh on Android (the iOS widgets fetch `/today` hourly with the stored JWT). `updatePeriodMillis` (30 min) just re-renders the cached snapshot so theming and battery staleness stay honest when the app hasn't run. Happy to add a WorkManager-free self-refresh in a follow-up if you want parity.
- The sleep value auto-sizes (`autoSizeTextType`) so "7h 17m" shrinks into its ring — the RemoteViews equivalent of the Swift `minimumScaleFactor(0.6)`.
- Tap anywhere on either widget opens the app.

## Testing

- `flutter analyze` clean for the touched files; full `flutter build apk --debug` compiles (their first CI compile would otherwise be at tag time, since build.yml only triggers on version tags).
- Not yet verified on a physical launcher — the preview above is a faithful rendering of the layouts, but I'd value a quick on-device check of the 2×2 sizing across OEM launchers before merge.

## Summary by CodeRabbit

* **New Features**
  * Added Android home-screen widgets displaying readiness, strain, sleep, and HRV metrics.
  * Added a band battery widget showing battery level, charging status, and last-connected name.
  * Added compact and expanded widget layouts with responsive sizing.
  * Added light and dark widget themes with automatic updates when theme or battery data changes.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -178,6 +178,31 @@
             </intent-filter>
         </receiver>
 
+        <!-- Home-screen widgets (Android siblings of ios/OpenStrapWidget). The app
+             pushes snapshots via home_widget (WidgetService); these just render. -->
+        <receiver
+            android:name=".OpenStrapWidgetProvider"
+            android:exported="true"
+            android:label="@string/widget_openstrap_label">
+            <intent-filter>
+                <action android:name="android.appwidget.action.APPWIDGET_UPDATE" />
+            </intent-filter>
+            <meta-data
+                android:name="android.appwidget.provider"
+                android:resource="@xml/widget_openstrap_info" />
+        </receiver>
+        <receiver
+            android:name=".OpenStrapBatteryWidgetProvider"
+            android:exported="true"
+            android:label="@string/widget_battery_label">
+            <intent-filter>
+                <action android:name="android.appwidget.action.APPWIDGET_UPDATE" />
+            </intent-filter>
+            <meta-data
+                android:name="android.appwidget.provider"
+                android:resource="@xml/widget_band_battery_info" />
+        </receiver>
+
         <!-- Don't delete the meta-data below.
              This is used by the Flutter tool to generate GeneratedPluginRegistrant.java -->
         <meta-data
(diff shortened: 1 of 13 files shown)


abdulsaheel:
great pr, thanks.

PR #97, opened 2026-07-19
feat(edge): detect + speak WHOOP 5 (gen5) alongside WHOOP 4

abdulsaheel:
## Summary
Teaches the BLE engine to detect and talk to WHOOP 5 (gen5 / \"fd4b\") bands as well as WHOOP 4. Pairs with OpenStrap/protocol#16 (the `BandProfile` abstraction). Analytics needs **no** changes — WHOOP 5 just feeds the same `decoded_onehz`/`decoded_rr` tables.

## Changes (all in `lib/ble/ble_engine.dart`)
- **Scan** filters both service UUIDs (gen4 `6108xxxx` / gen5 `fd4bxxxx`).
- **Discovery** pins the generation per session (`_Session.applyBand`) and rebuilds the frame reassemblers with the matching header shape; characteristics resolved via the generation's GATT prefixes.
- **`BandProfile` threaded** through the frame path, the command builder (`_send`), and the **history-result ACK** (safe-trim token echo).
- **Gen5 handshake** branch in `sendInit`: client-hello (`0x91`) + empty-payload `GET_DATA_RANGE`/`SEND_HISTORICAL_DATA`.
- **Record ingest** routes gen5 through `parseGen5Record`; unknown/motion kinds fall through to `raw_archive` as before.

## Testing
- `flutter test --concurrency=1` → **483 pass** (no regression).
- `flutter analyze lib/ble/` → clean.

## Invariants preserved
- The WHOOP 4 path is byte-identical (gen4 is the default profile).
- Safe-trim ACK invariant unchanged: durable commit still precedes the (now band-correct) ACK token echo.
- Live high-rate streams still never persisted; `decoded_*` schema unchanged.

## ⚠️ Known limitation
The **gen5 connect/handshake path is not yet validated on physical hardware** (no WHOOP 5 band available) — clearly marked in-code. Decode is validated against real owned captures + synthetic vectors. Requires an on-device pairing pass (bond → handshake → offload → ACK → trim → decode → day derive) before shipping to gen5 users.

## Summary by CodeRabbit

* **New Features**
  * Added WHOOP 5 pairing, device selection, and generation display.
  * Added WHOOP 5 historical data downloads, optional deep-buffer support, and live IMU streaming.
  * Added generation-aware clock, alarm, haptic, and data-transfer controls.
* **Bug Fixes**
  * Improved Bluetooth discovery, session setup, acknowledgments, and historical transfers.
  * Prevented invalid sensor data from affecting sleep detection and alarm state.
* **Chores**
  * Updated WHOOP 5 protocol support and app version.


abdulsaheel:
Addressed in 67fa1cb — extracted `_sendGetDataRange`/`_sendHistoricalData` helpers that pick the payload by generation (gen4 `[0x00]`, gen5 empty) and routed the init handshake, periodic backfill, manual refresh, and retry paths through them, so the gen5 offload command format is now identical everywhere. `flutter analyze` clean, 483 tests still green.

@coderabbitai review

Aurelian-Shuttleworth:
I have a whoop 5 that I am willing to use to validate against if that would help?

majidtahir1:
I was able to connect with my whoop 5.  so far i can see heart rate data live.  I will keep it connected to see recovery/strain and other data points.

abdulsaheel:
> I was able to connect with my whoop 5. so far i can see heart rate data live. I will keep it connected to see recovery/strain and other data points.

Great !! Thanks, lets continue this discussion here: https://github.com/OpenStrap/edge/discussions/178

abdulsaheel:
@coderabbitai full review

Requesting a full pass rather than incremental: this branch has accumulated three merged PRs since your last review on 08-01, and merging PRs *into* a PR's head branch doesn't reliably fire the `synchronize` event, so none of them have been reviewed as assembled.

What landed since you last looked:

- **#188** — gen5 SET_CLOCK revision byte + honest v18 lenient decode. Also a P0 fix: absent gravity was persisted as `(0,0,0)`, and `zAngle(0,0,0)` is exactly `0.0`, so a run of missing accel scored as *perfect immobility* and fabricated a ~7.9 h staged sleep window. Guarded via `Substrate.accelPresentAt` + a van Hees coverage floor.
- **#194** — WHOOP 5 wake alarm (slot index 1 + Maverick test buzz).
- **#195** — Maverick live IMU arm/decode for step calibration.

Two notes on your existing findings, so the pass doesn't re-litigate them:

1. **"Reuse gen5 empty-payload commands for later refreshes"** — I believe this is resolved. `_offloadPayload` is generation-aware and `_sendGetDataRange`/`_sendHistoricalData` both route through it, with `_startHistoricalRefresh` calling those helpers, so every trigger emits the gen5 form on a gen5 link. Worth confirming rather than taking my word for it.
2. **"Update the stale protocol-generation comment"** — still valid, and now more so. `pubspec.yaml` lines 30–31 claim this branch is WHOOP-4-only with protocol on `main`, contradicting the pin below it. Since then **protocol#16 merged**, so the pinned `c525e29` is now reachable from protocol `main` (`367d22b`) — meaning the current pin is a PR-branch head on a branch that gets deleted when this PR merges, which the same file explicitly warns against.

Areas I'd especially value a look at, given this is the point where gen5 becomes user-facing:

- the two derive/ingest paths and whether gen5 records reach `decoded_onehz` consistently
- anything where absent gen5 data could read as a measurement rather than an absence
- the alarm payload construction and its confirmation path

abdulsaheel:
@coderabbitai — worked through all five. **Four were real and are fixed in `9fd17b9`**; one I've declined with reasoning. No merge/conflict work in this pass.

### ① Vacuous guard test (Major) — valid, and it was my own

You were right: the file only exercised `accelPresentFraction` and never the gate it feeds. The coverage floor could have been deleted and every assertion would still have passed — the same vacuity I'd flagged in *other people's* tests earlier in this stack. It now drives `calendarDays` and asserts on `PhysioDay.sleepSource`.

Two things worth recording from building it, because they'd bite anyone repeating it:

- The first **positive control failed** — a "near-still" night wasn't detected as sleep at all, which would have made the negative test pass for the wrong reason. Fixed by giving the fixture real shape: constant gravity through the night, a 10°/s ramp while awake. It has to be a *ramp*, not an alternation — the mask smooths z-angle with a 5-second rolling **median**, which erases a 1 Hz square wave entirely.
- Mutation testing then showed the **all-zero case still wasn't discriminating**: with the gate removed it passed anyway, because a record immobile end-to-end is rejected downstream regardless. The discriminating shape is the realistic one — evening has gravity, the night's records don't. Removing the gate now fails **2** tests. I kept the all-zero case but labelled it as documentation rather than a guard.

### ② Stale serial (Minor) — valid

Confirmed the whole chain: `select` compares with `==`, `DeviceState` declares no `==`/`hashCode`, and `BleEngine` mutates `state.serial` in place — so the selector returned the same reference and no rebuild fired. Now selects the serial string, which is the only value this sheet reads from `device`/`paired`.

### ③ Chip overflow (Minor) — valid, `Wrap` applied

### ④ Persistence coverage (Minor) — valid

The R10-lite case asserted only *absence* from `decoded_onehz`, which would equally pass if the record were dropped outright. Added the `samples` retention half, plus a case protecting the unparseable-hex fallback in `_decodeOneHzSample`.

### ⑤ `gen5DeepBuffersEnabled` unwired — accurate, but not a defect

Verified: zero overrides in `app_state` or `background_sync`. But it's a deliberate default-OFF opt-in for v20/v21/v26 buffers that are **archived, not interpreted**. Wiring a settings toggle for a feature nothing consumes would be premature. Declined.

---

### One you dropped, which is still live

Your 08-01 finding on `pubspec.yaml` didn't reappear in this review, but it hadn't been fixed. It's real and I've now corrected it — the file claimed this branch was *"intentionally WHOOP-4-only — protocol stays on main"* while pinning the multiband commit twelve lines below.

It's also become more serious since you filed it: **protocol#16 has landed**, so the pinned SHA is now reachable from protocol `main` (`367d22b`). The current pin is a PR-branch head on `feat/multiband-whoop5` — the branch this PR deletes on merge — which is precisely the orphaning the same paragraph warns against. I left the pin itself alone (it belongs with the main-merge, not a review pass) and recorded it as a follow-up in the file.

Flagging because silence from a re-review isn't clearance.

---

12 tests added, gate tests mutation-verified, `analyze` clean. Suite **1085 passing / 6 failing** — the 6 are `notification_dedupe` time bombs this branch still carries because it predates #207 on main (19 hardcoded dates, no `todayLabel()`). They clear when main is merged in.

PR #104, opened 2026-07-19
fix: app stuck forever on the launch screen (Samsung) — pre-runApp inits had no timeout

dannymcc:
## Symptom

A Samsung user reported the app "just doesn't load": the native launch screen (cream background + logo mark) shows forever — no crash, no ANR dialogue, no UI. Force-quit and relaunch doesn't help.

## Root cause

`main()` awaits ~ten platform-channel inits before `runApp` — Firebase, two WorkManager cancels, flutter_blue_plus options, then widgets/notifications/prefs/theme/units/coach-config. Every one is guarded against **throwing** (`try/catch` / `_safeInit`), but none has a **timeout**. A channel call that simply never completes blocks the first frame indefinitely: the platform thread stays responsive so there's no ANR, and nothing was thrown so nothing is logged. The app is silently bricked at the splash.

The strongest device-specific suspect is `CoachConfig.load()` — a flutter_secure_storage read, i.e. the Android Keystore, which has well-documented hang-forever failure modes on some Samsung Knox devices (typically after OS updates). It was also the very last await before `runApp`. The repo has prior art for this failure class: the comment above the WorkManager cancels describes an earlier "hang on the loading screen" bug from this same phase.

## Fix

- **6 s ceiling on every pre-runApp init** — the inline awaits (Firebase, WorkManager, FBP) and everything routed through `_safeInit`. A timed-out init logs `[main] X init TIMED OUT after 6s — continuing without it` and the app boots degraded instead of not at all. The log line also turns any future report of this class into a one-look diagnosis.
- **`CoachConfig.load()` is no longer awaited at all** — nothing needs the AI-coach config until an AI screen opens, and `CoachConfig` is a `ChangeNotifier`, so consumers rebuild whenever the load lands. The keystore can now never gate the first frame. (Still bounded at 15 s so a wedged read doesn't hold the future forever.)

Degradation semantics are unchanged from what the existing try/catches already accepted: a failed *or slow* optional init proceeds with defaults (theme falls back to system, prefs to defaults, etc.).

## Verification

Field-tested on the affected Samsung device: a side-by-side debug build of this branch **boots and runs normally on the same phone where the production build never leaves the launch screen**. That confirms the hung-init mechanism (a rendering-layer problem would have hung the test build identically). The tester's build didn't capture logcat, so which specific plugin hangs on that device isn't pinned down — but with this patch any occurrence logs the culprit by name, so the next report answers that question by itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app startup reliability by preventing long-running initialization tasks from hanging indefinitely.
  * Time-limited theme and unit bootstrapping now fall back safely if they fail or time out.
  * Coach configuration loading no longer blocks the first screen; it runs in the background with timeout handling.
  * Added clearer timeout detection and logging while ensuring the app continues launching.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -17,17 +17,26 @@ import 'widget/widget_service.dart';
 import 'package:firebase_core/firebase_core.dart';
 import 'firebase_options.dart';
 import 'package:workmanager/workmanager.dart';
+import 'dart:async';
 import 'dart:io';
 import 'compute/background_derivation.dart' show kHeavyDeriveTaskName, kSyncTaskName;
 
+/// Ceiling on every pre-runApp platform-channel await. Each one is guarded
+/// against THROWING, but a channel call that simply never completes (seen in
+/// the field: flutter_secure_storage wedging on some Samsung Knox keystores)
+/// used to park the app on the native launch screen forever — no crash, no
+/// ANR, just "the app doesn't load". A timed-out init logs and degrades;
+/// first frame always ships.
+const _kStartupInitTimeout = Duration(seconds: 6);
+
 Future<void> main() async {
   WidgetsFlutterBinding.ensureInitialized();
-  
+
   // Initialize Firebase (overridden by dummy values until flutterfire configure)
   try {
     await Firebase.initializeApp(
       options: DefaultFirebaseOptions.currentPlatform,
-    );
+    ).timeout(_kStartupInitTimeout);
   } catch (e) {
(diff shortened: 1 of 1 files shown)


dannymcc:
Took the Future.wait suggestion — the two cancels now share one 6s ceiling instead of stacking two.

abdulsaheel:
Thanks for the fix.

PR #108, opened 2026-07-20
fix(readiness): stop duplicate-day pollution blanking the readiness ring

### LINKED ISSUE #106: Readiness keeps going blank
The readiness figure on the home screen keeps going blank for no reason

Image

dannymcc:
Fixes #106

## The bug

The Today dashboard intermittently renders a **blank `—` in the readiness ring** even though the score clearly exists — in the reported case the AI briefing banner on the same screen said *"Your readiness is solid at 74…"* while the ring below showed a dash. It looked random.

## Root cause

Readiness is a robust z-score composite scored against a rolling 28-day personal baseline, persisted as a `rolling_artifact` rebuilt from an in-memory `_BaselineHistoryCache` whose `appendScalars` **only appends, with no day identity**.

- Every BLE drain kicks a **light derive pass over _today_**, appending today's values and persisting the window.
- Re-deriving the same day **stacks duplicate copies of today** into the 28-slot window (evicting real history).
- Once ≳15 slots hold the same value, the composite's **MAD collapses to 0**, `robustZ` returns `null` for every input → the composite goes **absent** → the ring renders `—`. The banner meanwhile reads a cached AI-briefing snapshot from earlier, so the two disagree.
- The flicker: a heavy/finalize sweep appends enough _distinct_ days to flush the duplicates → score returns; light passes re-pollute → dash returns.

The codebase even anticipated this — the changelog note records a prior *"including tonight's own value in its own baseline"* bug; this is that failure mode reintroduced through the persisted artifact and amplified N-fold.

## The fix

Rebuild the persisted artifact from **`metric_series`** instead of the append-accumulated list. `metric_series` is keyed `(date, key)` with `REPLACE`, so it is **structurally one value per day** and cannot carry duplicate-day pollution. It **self-heals** already-polluted installs on the first derive after upgrade.

- New `LocalDb.trailingSeriesValues(key, n)` — the trailing (newest) `n` days, oldest→newest.
- `_refreshBaselines()` now rebuilds from that store on every persist.
- The in-memory **intra-sweep** path (`appendScalars` / `_attachHistory` feeding the live computation) is deliberately **unchanged** — only what gets persisted changes.

### Secondary bug fixed
The decode-failure fallback used `metricSeries(limit: 28)` — `date ASC LIMIT 28`, i.e. the **oldest** 28 days. Now uses the correct trailing window.

## Tests

`test/readiness_baseline_pollution_test.dart`:
1. **Mechanism** — a baseline dominated by one repeated value makes `robustZ` return `null` (MAD=0); a distinct baseline resolves.
2. **Fix** — re-deriving the same day 15× leaves exactly one row, so `trailingSeriesValues` stays de-duplicated and `robustZ` resolves; and it returns the newest `n` (oldest→newest), contrasted against the old leading-`n` behaviour.

## Verification note

Root cause verified end-to-end in code (Fable-assisted diagnosis, ~85%). To confirm on an affected device: the `rolling_artifact` series arrays will show long runs of one repeated value, and `readiness_absent` telemetry will show all values present + baselines full yet absent — only explicable by MAD degeneracy.

https://claude.ai/code/session_016q4RHdYkMAWuvDMFYNJbqp

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved readiness/illness baseline stability by rebuilding the rolling baseline window from canonical trailing per-day metric values, avoiding polluted stored rolling artifacts.
  * Fixed duplicate-day baseline pollution that could collapse the baseline and cause readiness indicators to disappear.
  * Updated baseline calculations and trailing-window behavior to use the newest N samples in chronological order (oldest → newest).
* **Tests**
  * Added regression coverage for polluted-baseline recovery, deduplication, trailing-window ordering, and ensuring polluted on-disk rolling artifacts are ignored.


dannymcc:
Good catch on the read path — fixed in 44cf351. `_BaselineHistoryCache.load()` now always rebuilds the window from the de-duplicated `metric_series` store rather than trusting the persisted `rolling_artifact`, so a valid-but-polluted artifact can't reach the computation on the first post-upgrade derive (or linger when every day is finalized and no refresh runs). Added a regression covering exactly that: a valid polluted artifact + all-finalized days, asserting the loaded window is clean and `robustZ` resolves.

abdulsaheel:
Nice work !

PR #109, opened 2026-07-20
fix(briefing): stop the briefing page rendering content twice

### LINKED ISSUE #107: Duplicates on the briefing page
There is multiple duplicates on the briefing page with Morning Briefing and User Safety being mentioned twice on the page

Image

dannymcc:
Fixes #107

## The bug

On the briefing page, content renders **twice** — the reporter's screenshot shows *"User Safety: safe"* as the hero headline **and** again as a bullet below it, plus *"Morning Briefing"* in both the screen title and the card overline.

## Root cause

Two independent duplications:

**A — content echoed (the real bug).** The user's BYOK model ignored the output format and returned a single line (`User Safety: safe`) instead of `one-liner / --- / bullets`. `parseBriefingResponse` then hit its empty-breakdown fallback:

```dart
if (rest.isEmpty) rest = '- $one';   // briefing_engine.dart
```

…which **copies the one-liner into the breakdown as a lone bullet**. `AiBreakdownScreen` renders the one-liner in the hero card *and* the breakdown card, so the same sentence appears twice.

**B — title echoed (cosmetic).** The hero card drew an overline `b.period.title.toUpperCase()` directly under the `AppScaffold` title, which is already `widget.period.title` — "Morning briefing" / "MORNING BRIEFING".

## The fix

- **`parseBriefingResponse`** now leaves `breakdownMd` **empty** when the model gave no distinct breakdown, instead of echoing the one-liner.
- **`AiBreakdownScreen`** hides the breakdown card when it is empty *or merely restates the one-liner* — so **already-cached** briefings (which stored the echoed `- one-liner`) also stop double-rendering, with no regeneration needed.
- **Drops the hero overline** (duplication B); the icon chip stays as the visual anchor.

Happy-path briefings (proper one-liner + bullets) are unchanged.

## Tests

Adds a parser regression: a single-line reply yields the one-liner with an **empty** breakdown (no echo). Existing separator/code-fence parse tests still pass (they have distinct bullet content, so they're unaffected).

https://claude.ai/code/session_016q4RHdYkMAWuvDMFYNJbqp

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented briefing summaries from showing the same sentence twice.
  * Improved handling for cached briefings so one-liners that were stored as single bullets don’t duplicate the hero text.
  * Breakdown markdown now renders only when it contains meaningful, non-duplicative content.
  * Removed the period title overline from the briefing hero display.
* **Tests**
  * Added parser coverage for single-line AI briefing responses.
  * Added widget coverage to ensure cached briefings don’t render a breakdown card when it would duplicate content.

### DIFF
--- a/lib/ai/briefing_engine.dart
+++ b/lib/ai/briefing_engine.dart
@@ -204,7 +204,11 @@ String buildBriefingUserPrompt(
   one = one.replaceFirst(RegExp(r'^\s*[-*#>]+\s*'), '').trim();
   if (one.length > 200) one = '${one.substring(0, 199)}…';
   rest = rest.trim();
-  if (rest.isEmpty) rest = '- $one';
+  // A model that returns a single line (no bullets) has no distinct breakdown.
+  // Leave it EMPTY rather than echoing the one-liner back as a lone bullet —
+  // the breakdown card would then render the same sentence a second time
+  // (the "duplicates on the briefing page" bug). The UI hides an empty
+  // breakdown, so the one-liner shows exactly once.
   return (oneLiner: one, breakdownMd: rest);
 }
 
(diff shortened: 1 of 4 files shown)


dannymcc:
Added in 4f2ff35 — a widget test seeds a cached briefing whose `breakdownMd` is just `- <oneLiner>` and asserts the sentence renders once with no breakdown card (`GptMarkdown` absent), covering the cached-cleanup path directly.

abdulsaheel:
I don't think this was an actual bug; rather, it was an LLM response issue. Nevertheless, the PR handles such cases gracefully. Thanks for the PR!

PR #152, opened 2026-07-25
remove Awake and Consistency tiles from the sleep screen bento

abdulsaheel:
took out both BentoTile entries from the right column of the sleep summary bento.

left the underlying data/getters alone since they're both still used elsewhere on the same screen:
- `_awakeMin` is still used in the Efficiency tile's long-press detail text
- Consistency still has its own honest-gated row further down the screen (shows "Need a few more nights" when there isn't enough history yet)

just removed the duplicate summary tiles up top, plus the now-unused `reg` local in `_summaryBento`.

had to update one test (`workout_sleep_redesign_test.dart`) that was asserting CONSISTENCY showed up in this bento — flipped it to assert both AWAKE and CONSISTENCY are gone from there instead, so it pins the new behavior instead of going stale.

611/611 green, analyze clean.

## Summary by CodeRabbit

* **UI Updates**
  * Simplified the sleep summary layout by removing the separate “Awake” and “Consistency” tiles.
  * Awake time remains available in the Efficiency details.
  * Consistency information is still available in the detailed metrics list.

### DIFF
--- a/lib/ui/sleep/sleep_detail_screen.dart
+++ b/lib/ui/sleep/sleep_detail_screen.dart
@@ -649,7 +649,6 @@ class SleepNightContent extends StatelessWidget {
     final eff = _efficiency;
     final debt = _debtMin;
     final noDebt = debt == null || debt.round() <= 0;
-    final reg = _regularity;
 
     void info(String title, String body, [String? method]) =>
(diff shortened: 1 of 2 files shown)


abdulsaheel:
@coderabbitai review

PR #163, opened 2026-07-28
fix(crashlytics): close the four top 0.9.19/0.9.20 production issues

abdulsaheel:
### **User description**
Four production defects from the 0.9.19/0.9.20 Crashlytics dashboard, each one a foreground assumption or an unbounded read applied to a context that breaks it. No analytics output changes, so `kAlgoVersion` stays at **50**.

## 1. Import/export OOM — `java.lang.OutOfMemoryError`

`importFromDb` ran `src.query(t)` per table and then held the entire result live for the duration of the insert loop. sqflite serialises a whole result set into Java objects on the platform side **before** any of it crosses the channel, so importing another device's `decoded_onehz` (86,400 rows per day of history) materialised the whole table on the 256 MB Dalvik heap at once (`target footprint 268435456, growth limit 268435456`). The OOM surfaced on whichever thread allocated next — which is why it was reported against a BLE binder callback rather than the import.

Both the import and export paths now page on `rowid`:

- **Keyset, not `LIMIT/OFFSET`** — OFFSET re-scans the skipped prefix on every page, which is quadratic over a full history.
- **One transaction per page**, not per table. The whole-table transaction could only ever commit if the entire table fit in memory first, which is the bug.
- **The orphan guard is unchanged in spirit**: each `decoded_onehz` row's guard is still queued in the *same* transaction as the row it guards. The invariant is per-row, not per-table.
- **Interrupt-safe**: every write stays `INSERT OR REPLACE` keyed on the row's own identity, so a re-run converges to the same state.
- On export, each page's RR beats are pulled and written before the next page is read, so peak residency is one page of `decoded_onehz` plus its beats — not a whole day of both.

## 2. Unbounded RR fallback read

`decodedRrByCounterRange` has a degraded branch for when an endpoint row isn't in `decoded_onehz` (a prune, or an import's REPLACE + orphan-guard DELETE landing between the page read and the call). The caller's counters are then just a span — and because the strap's counter resets on reboot, a reboot-straddling page degenerates to `0 .. ~1200000`, i.e. effectively the whole table, on the same heap that OOMed above.

Now capped. A page is 2000 frames and a second rarely carries more than a handful of beats, so the cap sits orders of magnitude above any legitimate page — **reaching it means the degraded path is serving a range it was never meant to**. Because a short read here shows up downstream as understated HRV rather than as an error, it is never silent: `decodedRrFallbackTruncations` counts it for the Diagnostics screen. (`db.dart` takes no telemetry dependency — it has to run inside compute isolates.)

## 3. Background derive timeouts — `day_blocks_failed`

`TimeoutException: day-blocks computation timed out after 0:01:30`, reported only from inside `IosBgTask._run` — i.e. only ever in **background**. Two foreground assumptions were being applied to a context that breaks both:

| | Foreground | Background (OS-throttled) |
|---|---|---|
| Day-lanes | 3 — real speedup from spare cores | **1** — no spare cores to soak up; 3 lanes divide one budget three ways and make each day ~3× slower in wall-clock |
| Per-day budget | 90 s | **4 min** — wall clock stops tracking work once the OS throttles us |

Paired with a wall-clock timeout, the old tuning converted "3 days derived" into "3 days timed out". Serial lanes also cap peak memory at one day's substrate instead of three. `background: true` is set at **construction** on all four headless entries (iOS BGProcessingTask + BGAppRefreshTask, Android WorkManager, post-drain background sync) so a long-lived foreground engine can never inherit background tuning by accident.

## 4. Jank watchdog false positives

The watchdog gated on `FrameTiming.totalSpan` — vsync-start to raster-end, which includes time the engine did nothing, above all the gap across an app resume where a "frame" straddles however long the app sat in the background. The result:

```
Slow frame: 16358ms (build=0 raster=8)
```

A 16-*second* stutter in which the app did 8 ms of work. That single false-positive class became the top issue by impacted-user count on both platforms, burying real signal underneath it.

`JankPolicy` gates on **build + raster** — the two costs the app controls and the two a fix would actually move. The 700 ms threshold is unchanged but now means "the engine burned 700 ms on one frame" rather than "700 ms elapsed", which is orders of magnitude rarer and always actionable. `totalSpan` is still attached as context alongside a derived `idle_ms`, so a genuine scheduling-delay/thermal pattern stays visible — it just isn't a jank *report*.

## Also: ImportScreen double-tap crash

`_busy` only went true once an import was already **running**, i.e. after the picker returned — so while the native sheet was still opening the three option cards stayed live. The sheet takes a beat to appear on a cold platform channel, so users do tap twice; the second tap called `pickFiles` again and file_picker threw `PlatformException(already_active)` out of an unawaited context, crashing the app. Guarded twice over: a `_picking` flag makes the cards inert for the whole picker lifetime, and the catch refuses to let any picker platform failure become fatal (`already_active` itself is silent — the first picker is still up and will deliver the user's choice).

## Testing

- `flutter test --concurrency=1` → **989 passed** (up from 914; three new suites: `db_paged_import_export_test`, `derive_pacing_test`, `jank_policy_test`).
- `flutter analyze` → clean, no issues.
- The tuning and verdict decisions live in pure classes (`DerivePacing`, `JankPolicy`) so they are unit-testable without a database, an isolate, a real background slot, a `SchedulerBinding`, or a Firebase app.
- Sibling pins untouched and verified: `pubspec.yaml` refs match `pubspec.lock` `resolved-ref`s (protocol `a98cd70`, analytics `f5ccae6`), git provenance intact.

Not yet exercised on real hardware — items 3 and 4 are background/timing behaviours that only prove out on a device over a few OS-granted slots, so the Crashlytics rates for `day_blocks_failed` and `jank_watchdog` are what should confirm this after release.

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Fix import/export OOM: replace unbounded `SELECT *` with keyset-paged reads (2000 rows/page, one transaction per page) on both import and export paths

- Fix background derivation timeouts: introduce `DerivePacing` so background slots use 1 lane and a 4-minute per-day timeout instead of 3 lanes and 90 s

- Fix false-positive jank reports: gate the watchdog on build+raster work time, not `totalSpan` (which includes app-resume idle gaps)

- Fix double-tap `already_active` crash in ImportScreen: add `_picking` guard around the native file picker call

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["ImportScreen\n(double-tap guard)"] -- "_picking flag\nprevents re-entry" --> B["FilePicker\n(platform channel)"]
  C["importFromDbFile\n/ exportDaysDb"] -- "keyset-paged\n2000 rows/txn" --> D["decoded_onehz\n/ decoded_rr\n(SQLite)"]
  E["DerivationEngine\n(background: bool)"] -- "delegates to" --> F["DerivePacing\n(concurrency + timeout)"]
  F -- "background=true\n→ 1 lane, 4 min" --> G["IosBgTask\n/ WorkManager\n/ bgsync-derive"]
  F -- "background=false\n→ ≤3 lanes, 90 s" --> H["Foreground derive"]
  I["FrameTiming\n(totalSpan)"] -- "replaced by\nbuild+raster" --> J["JankPolicy\n→ JankVerdict"]
  J -- "non-fatal only\nwhen work≥700ms" --> K["Crashlytics"]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><details><summary>8 files</summary><table>
<tr>
  <td><strong>db.dart</strong><dd><code>Keyset-paged import/export; bounded RR fallback read</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-b96f416000a07a912224053a34f440c27993e014d324d6ae8c9dfe88bb0617e6">+200/-101</a></td>

</tr>

<tr>
  <td><strong>derivation_engine.dart</strong><dd><code>Wire background flag into concurrency and per-day timeout</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-0e401d8cdb1091d9a9591e9d92c4c5e979293f8480da203b22011a7cf9072ea6">+19/-11</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>background_derivation.dart</strong><dd><code>Pass background=true to DerivationEngine in WorkManager entry</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-6481db284babd77b9ea0a6c6a788a2c48efc5dcd8af8abb2f7a1878e2d8c8113">+2/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>background_sync.dart</strong><dd><code>Pass background=true to post-drain derive pass</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-16ac57f05bd5fda34d2e44e6b163a9053f857df64e384dd6b6ec95170d07dab9">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>ios_bg_task.dart</strong><dd><code>Pass background=true to both iOS BGTask derive engines</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-37f2f6bd5c40194b7817d4f0f614f7778403791c7f64fc26b3f955e1da0b3dae">+4/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>jank_policy.dart</strong><dd><code>New pure class: jank verdict based on engine work, not totalSpan</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-a7db4fc1af6b0b55ed145baa3af16431abd32d795604749ef13e038c9d3a3895">+90/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>telemetry_service.dart</strong><dd><code>Replace totalSpan jank trigger with JankPolicy build+raster check</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-1377388a2f920a4e598fa88333d0520d928c2dd5e181bed8c8a3b5e1e884b796">+30/-15</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>import_screen.dart</strong><dd><code>Guard file picker against double-tap already_active crash</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-1cd123af00fd08da6de79d52e4896953b639ccc8c05f6048adcb369e9447966f">+48/-10</a>&nbsp; </td>

</tr>
<tr>
  <td><strong>derive_pacing.dart</strong><dd><code>New pure class: foreground vs background lane/timeout policy</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-d077d596a83ebd1f903843e675e7eb97c248e421b8562dbc39943d61993feba5">+56/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>db_paged_import_export_test.dart</strong><dd><code>Tests: paged import/export correctness across page boundaries</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-4772c6a9fe59ee8fe243ecc394f23dd5338bb60cb1ed517a512d3cecb21f7a8c">+265/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>derive_pacing_test.dart</strong><dd><code>Tests: background vs foreground pacing concurrency and timeout</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-793a9acf5d53bee025d7a568199b2dec09d74dc5d234f6dea0ec3054605c2c8d">+66/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>jank_policy_test.dart</strong><dd><code>Tests: jank policy suppresses idle-span false positives</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-8e45277cb45d1386860cd99042c6b30b4761037b47968efe39287fe989f6dfd3">+90/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Configuration changes</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>AppStoreConnect.plist</strong><dd><code>Add App Store Connect export options plist for iOS archive</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/163/files#diff-d47033b230a04822c9a4fc95a0d102c50c1b03ac8070732976593a36fe1ed9b3">+18/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

## Summary by CodeRabbit

* **New Features**
  * Added adaptive compute pacing for foreground vs background derivations.
  * Improved jank detection to report engine work based on derived timing (work vs idle).
  * Added iOS App Store Connect export configuration.

* **Bug Fixes**
  * Improved reliability of heavy/background sync derivations on iOS.
  * Bounded RR fallback reads and added import/export paging to prevent excessive memory use and reduce incomplete data outcomes.
  * Prevented duplicate file-picker actions and improved picker error handling.

* **Tests**
  * Added regression coverage for pacing, database paging import/export integrity, and jank policy behavior.

### DIFF
--- a/ios/ExportOptions/AppStoreConnect.plist
+++ b/ios/ExportOptions/AppStoreConnect.plist
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>method</key>
+    <string>app-store-connect</string>
+    <key>teamID</key>
+    <string>2U62X3RF3R</string>
+    <key>signingStyle</key>
+    <string>automatic</string>
+    <key>destination</key>
+    <string>upload</string>
+    <key>uploadSymbols</key>
+    <true/>
+    <key>stripSwiftSymbols</key>
+    <true/>
+</dict>
+</plist>
--- a/lib/compute/background_derivation.dart
+++ b/lib/compute/background_derivation.dart
@@ -66,7 +66,8 @@ void derivationDispatcher() {
       } else if (task == kHeavyDeriveTaskName) {
         debugPrint('[bg-derive] triggered by WorkManager');
         final profile = await _loadProfile();
-        final engine = DerivationEngine(log: (m) => debugPrint('[bg-derive] $m'));
+        final engine = DerivationEngine(
+            log: (m) => debugPrint('[bg-derive] $m'), background: true);
         await engine.run(profile, heavy: true);
         // Baseline-dirty rescan on the scheduled tick: refresh baseline-dependent
         // scalars on recent finalized days when the rolling baseline has moved.
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -37,6 +37,7 @@ import '../notify/notification_event.dart';
 import '../notify/tap_router.dart' show kRouteWorkoutSuggestion;
 import '../telemetry/telemetry_service.dart';
 import 'crossday_pipeline.dart';
+import 'derive_pacing.dart';
 import 'derive_prepare.dart';
 import 'onehz_pipeline.dart';
 import 'profile.dart';
@@ -655,8 +656,16 @@ Future<void> runWithConcurrency<T>(
 }
 
 class DerivationEngine {
-  DerivationEngine({this.log});
+  DerivationEngine({this.log, this.background = false});
   final void Function(String)? log;
+
+  /// True when this engine was constructed inside a headless/background entry
+  /// (iOS BGProcessingTask / BGAppRefreshTask, Android WorkManager, the
+  /// post-drain background sync pass). The OS throttles CPU hard in those
+  /// contexts, which changes two tuning decisions — see [_deriveConcurrency]
+  /// and [_perDayTimeout]. Set at construction, not per-run, so a long-lived
+  /// foreground engine can never inherit background tuning by accident.
+  final bool background;
 
   bool _running = false;
   bool get running => _running;
@@ -1493,9 +1502,13 @@ class DerivationEngine {
         'v$kAlgoVersion|na';
   }
 
+  /// Foreground vs background pacing — lane count and per-day wall-clock
+  /// budget. See [DerivePacing] for why the background numbers differ.
+  DerivePacing get _pacing => DerivePacing(background: background);
+
   /// Max wall-clock for ONE day's off-isolate compute. On timeout the day is
   /// skipped so the sweep always makes progress.
-  static const Duration _perDayTimeout = Duration(seconds: 90);
+  Duration get _perDayTimeout => _pacing.perDayTimeout;
 
   /// Throttle for the readiness-absent diagnostic log — one per calendar day
   /// so repeated light-pass re-derives of today don't spam the outbox.
@@ -1510,17 +1523,12 @@ class DerivationEngine {
   /// substrate loads + compute-isolate all finishing before the next day even
   /// started), which wastes every core beyond the one doing the current day's
   /// work. Running several days' isolate work genuinely concurrently gets
-  /// real wall-clock speedup from the device's other cores. Capped
-  /// conservatively — this is a phone doing background/foreground compute,
-  /// not a server batch job — rather than using every available core.
-  static const int _maxDeriveConcurrency = 3;
-
+  /// real wall-clock speedup from the device's other cores — in the FOREGROUND.
+  /// A headless background slot has no spare cores to soak up, so it takes one
+  /// lane; [DerivePacing] owns that decision and explains it.
   int get _deriveConcurrency {
     try {
-      return math.max(
-        1,
-        math.min(_maxDeriveConcurrency, Platform.numberOfProcessors),
-      );
+      return _pacing.concurrency(Platform.numberOfProcessors);
     } catch (_) {
       return 1; // Platform unavailable on this target — sequential fallback
     }
--- a/lib/compute/derive_pacing.dart
+++ b/lib/compute/derive_pacing.dart
@@ -1,0 +1,56 @@
+// derive_pacing.dart — how hard to push per-day derivation, and how long to
+// wait for it, depending on whether we are in the foreground or in a headless
+// OS-granted background slot.
+//
+// WHY THIS EXISTS (production, Crashlytics 0.9.20 / iOS 27):
+// `_runDayBlocksCancellable` was reporting `day_blocks_failed` —
+// "TimeoutException: day-blocks computation timed out after 0:01:30" — from
+// inside `IosBgTask._run`, i.e. only ever in BACKGROUND. Two foreground
+// assumptions were being applied to a context that breaks both:
+//
+//  1. CONCURRENCY. Running 3 day-lanes concurrently is a win when there are
+//     spare cores. A background task does not get spare cores — it gets a
+//     throttled slice of CPU. Three lanes therefore do not go faster; they
+//     divide one budget three ways and make each day ~3x slower in wall-clock.
+//     Paired with a wall-clock timeout, that converts "3 days derived" into
+//     "3 days timed out". Serial lanes also cap peak memory at one day's
+//     substrate instead of three.
+//
+//  2. TIMEOUT. The 90 s guard exists to survive a HUNG day, but it is measured
+//     in wall clock, and wall clock stops tracking work once the OS throttles
+//     us. A day that computes in 20 s foreground can legitimately need several
+//     times that in a BGProcessingTask on a busy or thermally-limited device.
+//
+// Kept pure and separate so the tuning is unit-testable without a database, an
+// isolate, or a real background slot.
+
+/// Pacing decisions for one derivation run.
+class DerivePacing {
+  const DerivePacing({required this.background});
+
+  /// True for headless entries: iOS BGProcessingTask / BGAppRefreshTask,
+  /// Android WorkManager, and the derive pass that follows a background drain.
+  final bool background;
+
+  /// Upper bound on foreground day-lanes. Deliberately conservative — this is
+  /// a phone doing work alongside the UI, not a server batch job.
+  static const int maxForegroundConcurrency = 3;
+
+  static const Duration foregroundPerDayTimeout = Duration(seconds: 90);
+  static const Duration backgroundPerDayTimeout = Duration(minutes: 4);
+
+  /// Worker-pool size. [cores] is the device's processor count; pass whatever
+  /// `Platform.numberOfProcessors` reported (callers that cannot read it should
+  /// pass 1 and get the sequential fallback).
+  int concurrency(int cores) {
+    if (background) return 1;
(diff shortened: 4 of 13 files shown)


abdulsaheel:
**Fixed** (8994285)

- **CodeRabbit, blanket `catch (_)`** — valid, and the best catch of the review. Narrowed to `DatabaseException.isNoSuchTableError()` with everything else rethrowing. Used sqflite's own predicate rather than `toString().contains('no such table')`, which isn't stable across platforms. Added a test: the fixture only creates `decoded_onehz` + `decoded_rr`, so every other table already exercises the missing-table path on every import test — meaning if that predicate ever stops matching, partial imports would start *throwing* rather than skipping, riding in on tests that look like they're about paging. Asserted by name now.
- **PR Agent, `_rowid` reaching `onPage`** — agreed it was latent rather than live, but the raw-row API was an open invitation, so `onPage` now receives rows with the cursor column already stripped.

**Declining, with measurements**

- **CodeRabbit, filtered rowid paging (🟠 Major)** — the query-plan claim is correct and I verified it (`SEARCH ... USING INTEGER PRIMARY KEY (rowid>?)` vs the `rec_ts` index on your proposed shape). But the severity rests on "a multi-year ledger", and that premise doesn't hold for this table: `rawRetentionDays = 3` (`derivation_engine.dart:356`) prunes `decoded_onehz` behind the data edge, so it holds days, not years. The never-pruned tables walked per-day by the `for (final dayId in sorted)` loop are hundreds to thousands of rows. Measured on a real 435k-row ledger, the worst case — the exhaustion page that scans to the end of the table — is **~10 ms** (vs ~12 ms for the index-driven shape). Per-table cursors would need a composite `(ts, rowid)` key for the non-unique columns; that's real complexity on a path that exists to fix an OOM, for no measurable gain. Tradeoff and revisit condition are now documented on `copyPaged`.
- **PR Agent, static counter invisible across the isolate boundary** — doesn't hold. The only caller is `derivation_engine.dart:1270`, which reads on the **main isolate** and ships each page to the compute worker via `worker.send`; every sqflite call needs the root isolate's platform channel regardless. Increments land in the same isolate that reads them. That said, the comment it was reasoning from was misleading, so I've rewritten it to state why a static is sound *here specifically* and what would break it.
- **PR Agent, confirm the export still writes `version`** — confirmed present: `db.dart:2988`, `version: schemaVersion`, with the comment explaining that its absence is exactly why this export path had never once produced a file (fixed in #158).
- **PR Agent, verify no engine reuses the flag across contexts** — verified, and it's correct by construction: `app_state.dart:122` holds the long-lived foreground engine and never passes `background:`, so it defaults to false; only the four headless entries pass true. That's why the field is `final` and set at construction rather than per-run.
- **PR Agent, Team ID in `ExportOptions`** — agreed with your own assessment that it isn't a credential. The signing material (`Signing.xcconfig`) stays gitignored.

990 tests pass, `flutter analyze` clean.

PR #172, opened 2026-08-01
fix bugs

abdulsaheel:
### **User description**
- stop dropping naps that straddle midnight
- unfreeze live step count on today
- don't cache a morning briefing off an unsettled overnight
- hold off deriving days next to finalized ones after a tz jump
- don't feed today's unfinalized night into the illness alert
- don't let a partial derive pass write into the baseline series
- hold steps display through a disconnect, checkpoint against app kill
- notify per workout bout, not once per day
- auto-retry an unconfirmed alarm once before warning

___

### **PR Type**
Bug fix

___

### **Description**
- Fix naps straddling midnight being silently dropped by introducing a `napSub` buffer window

- Fix live step count freezing/dropping on disconnect with a grace cushion and orphan-session recovery

- Fix false illness alerts, stale morning briefings, and partial-derive baseline pollution

- Fix alarm confirmation retry, per-bout workout notifications, and timezone-travel derive guard

___

### Diagram Walkthrough

```mermaid
flowchart LR
  napSub["napSub buffer\n(+3h past midnight)"]
  napDetect["Nap detection\n(_sleepPeriods / _attachNaps)"]
  napSub -- "attributionEndSec guard" --> napDetect

  liveSteps["liveSteps getter"]
  cushion["_sessionStepsCushion\n(20s grace)"]
  checkpoint["_checkpointLiveSession\n(SharedPrefs, per-minute)"]
  orphan["_recoverOrphanedLiveSession\n(on connect)"]
  liveSteps -- "max(raw, cushion)" --> cushion
  liveSteps -- "checkpoint on commit" --> checkpoint
  checkpoint -- "recovered on next connect" --> orphan

  briefing["Morning briefing\ngeneration"]
  overnightState["overnight_state == 'ready'\ncheck"]
  briefing -- "gate on" --> overnightState

  tzGuard["_timezoneTravelSuspected\n(>=3h jump)"]
  deriveScope["_buildDeriveScope\npending days"]
  tzGuard -- "hold adjacent days" --> deriveScope

  partialDerive["partial derive pass"]
  metricSeries["metric_series write"]
  partialDerive -- "skipped when partial=true" --> metricSeries

  illnessAlert["Illness/anomaly CUSUM"]
  todayRow["today's unfinalized row"]
  todayRow -- "excluded from" --> illnessAlert

  alarmRetry["_onAlarmGraceElapsed\nauto-retry once"]
  alarmConfirm["AlarmConfirmation\nevent 56"]
  alarmRetry -- "silent re-arm before warning" --> alarmConfirm

  workoutNotif["auto_workout notification"]
  boutKey["dedupeKey: date:auto_workout:endSec"]
  workoutNotif -- "per-bout key" --> boutKey
```

<tr>
  <td>
    <details>
      <summary><strong>derive_prepare.dart</strong><dd><code>Add napSub buffer slice to fix midnight-straddling nap bisection</code></dd></summary>
<hr>

lib/compute/derive_prepare.dart

<ul><li>Adds <code>napSub</code> field to <code>PreparedDerivationDay</code>: a substrate slice extended <br>3 hours past the calendar day end so naps straddling midnight are seen <br>whole<br> <li> Introduces <code>napBoundaryBufferSec</code> constant (3 h) documenting the buffer<br> <li> Serializes/deserializes <code>napSub</code> in <code>toJson</code>/<code>fromJson</code>; falls back to <br><code>daySub</code> for old rows<br> <li> Passes <code>napSub = sub.slice(day.startSec, day.endSec + </code><br><code>napBoundaryBufferSec)</code> in <code>prepareDerivationPayload</code></ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/172/files#diff-1d7f803b5a943fdcb7d6357773a0f329125e028695258013ec44b5eeec7f4dcc">+27/-4</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>derivation_engine.dart</strong><dd><code>Fix nap double-count, TZ guard, illness alert, and per-bout workout </code><br><code>notify</code></dd></summary>
<hr>

<ul><li>Threads <code>napSub</code> through <code>_DayBlocksInput</code> and uses it (with <br><code>attributionEndSec</code>) in <code>_sleepPeriods</code> and <code>_attachNaps</code> so naps starting <br>in the buffer but belonging to tomorrow are excluded<br> <li> Adds <code>_timezoneTravelSuspected()</code> (>=3 h offset jump) and holds pending <br>days adjacent to finalized ones until a full restage; resets guard on <br>force-restage<br> <li> Excludes today's unfinalized <code>day_result</code> row from the cross-day <br>illness/anomaly CUSUM feed<br> <li> Changes workout notification <code>dedupeKey</code> from <code>date:auto_workout</code> to <br><code>date:auto_workout:endSec</code> so a second real workout the same day is not <br>silently swallowed<br> <li> <code>_adjacentDayIds</code> helper uses DST-safe <code>DateTime</code> arithmetic</ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/172/files#diff-0e401d8cdb1091d9a9591e9d92c4c5e979293f8480da203b22011a7cf9072ea6">+120/-10</a></td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>db.dart</strong><dd><code>Skip metric_series writes for partial derive passes</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

lib/data/db.dart

<ul><li>Wraps <code>metric_series</code> inserts in <code>upsertDayResult</code> with <code>if (!partial)</code> <br>guard so a failed/timed-out second-half derive pass cannot overwrite <br>or seed baseline series values</ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/172/files#diff-b96f416000a07a912224053a34f440c27993e014d324d6ae8c9dfe88bb0617e6">+14/-6</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>app_state.dart</strong><dd><code>Fix step freeze, step checkpoint, briefing cache, and alarm auto-retry</code></dd></summary>
<hr>

lib/state/app_state.dart

<ul><li>Adds <code>_sessionStepsCushion</code> / <code>_sessionCushionSetAtMs</code> (20 s grace) to <br><code>liveSteps</code> so the displayed count never regresses on disconnect before <br>derivation catches up<br> <li> Adds <code>_checkpointLiveSession</code> (writes committed steps to SharedPrefs <br>once per minute) and <code>_recoverOrphanedLiveSession</code> (replays checkpoint <br>on next connect) to survive process kills<br> <li> Gates morning briefing generation on <code>overnight_state == 'ready'</code> to <br>avoid caching a briefing off a still-syncing overnight<br> <li> Adds <code>_alarmAutoRetried</code> flag and <code>_onAlarmGraceElapsed</code> to silently <br>re-arm the alarm once before surfacing the unconfirmed warning<br> <li> Adds <code>liveSteps</code> to the <code>context.select</code> tuple in <code>TodayScreen</code> so the steps <br>tile updates during a live walk<br> <li> Calls <code>_recoverOrphanedLiveSession()</code> before <code>_resetLivePedometer()</code> on <br>connect; clears checkpoint after clean session end</ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/172/files#diff-d56ba858e512e18dc3962da198adee46d354aca39ff7c1b4a11d5f19b9afc1ae">+168/-4</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>briefing_engine.dart</strong><dd><code>Remove stale time-of-day greeting from cached briefing prompt</code></dd></summary>
<hr>

lib/ai/briefing_engine.dart

<ul><li>Removes time-of-day greeting from the system prompt entirely; <br>instructs the model to start with substance so a cached briefing read <br>hours later doesn't open with a stale greeting<br> <li> Updates <code>partOfDay</code> doc comment to clarify it is generation-time context <br>only, not a greeting directive</ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/172/files#diff-0dbec32919b228837406641ca5e1d2a11b648ae9c5291fd0ba436b715f7798ce">+11/-6</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>today_screen.dart</strong><dd><code>Include liveSteps in Today screen select to unfreeze step tile</code></dd></summary>
<hr>

lib/ui/today/today_screen.dart

<ul><li>Adds <code>a.liveSteps</code> to the <code>context.select</code> tuple so the Today screen <br>rebuilds on live step changes (~1/s) without waiting for an unrelated <br><code>dbCounts</code> change</ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/172/files#diff-98054aa9fb73d26ff39f204a6cc09a01189e143e79dd1f4a2c36c941a8d8c987">+10/-10</a>&nbsp; </td>

</tr>

</details>

___

## Summary by CodeRabbit

* **New Features**
  * Live step counts now update more reliably on the Today screen.
  * Interrupted step-tracking sessions can recover automatically.
  * Alarm confirmation includes an automatic silent re-arm.
  * Nap detection better handles periods crossing calendar-day boundaries.
  * AI briefings display greetings separately and begin directly with briefing content.

* **Bug Fixes**
  * Prevented incomplete daily results from overwriting baseline metrics.
  * Improved duplicate workout notifications and timezone-change handling.
  * Excluded unsettled daily data from illness and anomaly alerts until finalized.
  * Improved handling of daylight-saving transitions and cross-day calculations.
  * Improved reliability when processing sleep and activity data.

### DIFF
--- a/lib/ai/briefing_engine.dart
+++ b/lib/ai/briefing_engine.dart
@@ -166,18 +166,17 @@ String readinessBand(num v) {
   return 'good';
 }
 
-String briefingSystemPrompt(BriefingPeriod period, String timeOfDay) {
+String briefingSystemPrompt(BriefingPeriod period) {
   final scope = period == BriefingPeriod.morning
       ? 'last night\'s sleep and recovery, and what they mean for the day ahead'
       : 'today\'s activity, strain and stress, and how the day landed';
   return 'You write a health briefing for a local-first fitness band app. '
       'Summarize $scope.\n'
       'HARD RULES:\n'
-      '- Do NOT open with a greeting or any time-of-day reference ("good '
-      'morning", "this evening", etc.) — the app shows its own greeting '
-      'separately, computed at the moment the reader actually opens it, and '
-      'this text may be read hours after it was written. Start straight with '
-      'the substance.\n'
+      '- Do NOT open with a greeting or any reference to the time of day — '
+      'the app shows its own greeting separately, computed at the moment the '
+      'reader actually opens it, and this text may be read hours after it was '
+      'written. Start straight with the substance.\n'
       '- Use ONLY the numbers provided. Never invent, estimate or mention a '
       'metric that is not in the data. No medical advice or diagnosis.\n'
       '- If a "readiness" value is given, its parenthesized band label '
@@ -291,7 +290,7 @@ class BriefingEngine {
         (({required String system, required String user}) =>
             CoachEngine.completeText(
                 config: config, system: system, user: user)))(
-      system: briefingSystemPrompt(period, tod),
+      system: briefingSystemPrompt(period),
       user: buildBriefingUserPrompt(period, day, inputs, tod),
     );
     if (raw.trim().isEmpty) {
--- a/lib/ai/briefing_engine.dart
+++ b/lib/ai/briefing_engine.dart
@@ -134,10 +134,12 @@ Future<Map<String, dynamic>> collectBriefingInputs(
 
 // ── prompt building (PURE — unit-tested on sample data) ───────────────────────
 
-/// The reader's local time of day — used only for the briefing's greeting, and
-/// deliberately distinct from [BriefingPeriod]: the morning briefing (last
-/// night's sleep + recovery) is shown right up to 17:00, so a reader opening it
-/// in the afternoon must be greeted for the afternoon, never the "morning".
+/// The reader's local time of day at GENERATION time — passed into the prompt
+/// as context only (the model is told not to write a greeting off it; the app
+/// renders its own greeting fresh at read time, since a cached briefing can be
+/// read hours after it was generated). Deliberately distinct from
+/// [BriefingPeriod]: the morning briefing (last night's sleep + recovery) is
+/// shown right up to 17:00.
 String partOfDay(DateTime now) {
   final h = now.hour;
   if (h < 12) return 'morning';
@@ -169,10 +171,13 @@ String briefingSystemPrompt(BriefingPeriod period, String timeOfDay) {
       ? 'last night\'s sleep and recovery, and what they mean for the day ahead'
       : 'today\'s activity, strain and stress, and how the day landed';
   return 'You write a health briefing for a local-first fitness band app. '
-      'It is currently $timeOfDay for the reader — if you open with a greeting, '
-      'greet for the $timeOfDay and never assume a different time of day. '
       'Summarize $scope.\n'
       'HARD RULES:\n'
+      '- Do NOT open with a greeting or any time-of-day reference ("good '
+      'morning", "this evening", etc.) — the app shows its own greeting '
+      'separately, computed at the moment the reader actually opens it, and '
+      'this text may be read hours after it was written. Start straight with '
+      'the substance.\n'
       '- Use ONLY the numbers provided. Never invent, estimate or mention a '
       'metric that is not in the data. No medical advice or diagnosis.\n'
       '- If a "readiness" value is given, its parenthesized band label '
--- a/lib/compute/crossday_pipeline.dart
+++ b/lib/compute/crossday_pipeline.dart
@@ -47,20 +47,30 @@ Map<String, dynamic> buildCrossDayBundle(
     for (final d in days) _numOrNull(d['skin_temp_z']),
   ];
 
+  // A day flagged `unsettled` (today, still syncing / not finalized) is a
+  // truncated reading, not a physiological signal — it must not drive an
+  // illness/anomaly/temperature ALERT. It stays in `days` for everything else
+  // (readiness, RHR trend, load, sleep debt, `recent`), which is why this is a
+  // per-input null rather than dropping the row from the list.
+  final unsettled = <bool>[for (final d in days) d['unsettled'] == true];
+  T? settled<T>(int i, T? v) => unsettled[i] ? null : v;
+
   // ── illness CUSUM (NightSignal) on nightly RHR ─────────────────────────────
-  final illness = ana.illnessCusum(dates, rhrList);
+  final illness = ana.illnessCusum(dates, [
+    for (var i = 0; i < n; i++) settled(i, rhrList[i]),
+  ]);
 
   // ── multivariate anomaly {RHR↑,HRV↓,temp↑,resp↑} ───────────────────────────
   // Build one AnomalyFeatures per day (same length as dates). Days with all
   // features null still occupy a slot — the detector handles the nulls
   // internally (needs ≥2 present features tonight to compute a distance).
   final feats = <ana.AnomalyFeatures>[
-    for (final d in days)
+    for (var i = 0; i < n; i++)
       ana.AnomalyFeatures(
-        rhr: _numOrNull(d['rhr']),
-        hrv: _numOrNull(d['rmssd']),
-        temp: _numOrNull(d['skin_temp_z']),
-        resp: _numOrNull(d['resp_rate']),
+        rhr: settled(i, rhrList[i]),
+        hrv: settled(i, rmssdList[i]),
+        temp: settled(i, tempList[i]),
+        resp: settled(i, respList[i]),
       ),
   ];
   final anomaly = ana.multivariateAnomaly(dates, feats);
@@ -81,7 +91,9 @@ Map<String, dynamic> buildCrossDayBundle(
   final load = ana.ctlAtlTsb(dailyTrimp);
 
   // ── skin-temp illness flag (Smarr, cycle-aware) ────────────────────────────
-  final tempIllness = ana.tempIllnessFlag(dates, tempList);
+  final tempIllness = ana.tempIllnessFlag(dates, [
+    for (var i = 0; i < n; i++) settled(i, tempList[i]),
+  ]);
 
   // ── circadian: mid-sleep, free/work split, jetlag, chronotype, sleep debt ──
   // mid-sleep epoch = (onset+wake)/2; local clock-hours in [0,24) via mod-day.
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -1836,6 +1836,7 @@ class DerivationEngine {
       // sendable object (never `this`, `day`, or `bundle`).
       final blocksInput = _DayBlocksInput(
         daySub: daySub,
+        napSub: day.napSub,
         sleepSub: sleepSub,
         profile: profile,
         onsetSec: day.sleepOnsetSec,
@@ -3267,7 +3268,12 @@ class DerivationEngine {
 
   /// Sleep periods: the main sleep + any NAPS (still, on-wrist minute-runs ≥20
   /// min OUTSIDE the main window). Conservative — naps need sustained stillness.
-  static Map<String, dynamic> _sleepPeriods(Substrate s, int onsetSec, int offsetSec) {
+  static Map<String, dynamic> _sleepPeriods(
+    Substrate s,
+    int onsetSec,
+    int offsetSec, {
+    int? attributionEndSec,
+  }) {
     final periods = <Map<String, dynamic>>[];
     var totalAsleep = 0;
     if (offsetSec > onsetSec) {
@@ -3317,8 +3323,13 @@ class DerivationEngine {
           j++;
         }
         final lenMin = j - i;
-        if (lenMin >= 20) {
-          final start = keys[i] * 60, end = keys[j - 1] * 60 + 60;
+        final start = keys[i] * 60;
+        // A run STARTING at/after the real day boundary belongs to tomorrow's
+        // own (unbuffered) window — only count runs that started today.
+        final startsToday =
+            attributionEndSec == null || start < attributionEndSec;
+        if (lenMin >= 20 && startsToday) {
+          final end = keys[j - 1] * 60 + 60;
           periods.add({
             'is_main': false,
             'start': start,
@@ -3342,8 +3353,9 @@ class DerivationEngine {
     Map<String, dynamic>? scMap,
     Substrate s,
     int onsetSec,
-    int offsetSec,
-  ) {
+    int offsetSec, {
+    int? attributionEndSec,
+  }) {
     try {
       final n = s.length;
       if (n < 60) return;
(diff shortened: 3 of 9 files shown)


abdulsaheel:
Went through the bot findings:

- TZ guard resetting itself every call (real bug — pushed a fix, only bootstraps/updates the baseline when NOT jumped, so the hold actually persists until \`force\`/Re-analyze runs).
- Orphan recovery raced against pedometer reset (checked — the recovery path only reads the SharedPreferences checkpoint, not the in-memory counters, so no double-count was possible, but awaited it anyway instead of unawaited for clarity).
- Alarm retry latch not reset on failure — checked every call site that touches \`_alarmGraceTimer\`; nothing re-arms it except \`setAlarm()\` itself (which already resets the latch) and this function's own success branch. The described "second grace-elapsed path" isn't reachable as written, so left as-is.
- Day-label helper in orphan recovery — same inline pattern already used in \`_finalizeLivePedometer\` right above it; not touching that in this PR.

abdulsaheel:
Went through the remaining bot findings against current `main`+branch code — here's what was still open and is now fixed in 03a9980, plus what I'm declining and why.

**Fixed**

- **`liveSteps` cushion leaked into persistence/calibration** (CodeRabbit, Critical) — confirmed real. `_finalizeLivePedometer` and `finishStepCalibration` both read the cushioned getter, so a short session ending inside the prior session's 20 s grace window persisted the *prior* total again → double-counted `live_coverage` + a nonsense cadence into the calibration model. Added `_rawSessionSteps` and used it at both persistence sites; `liveSteps` is now display-only.
- **Cushion overwrite could regress the display** — same block, fixed: the cushion never moves down.
- **Orphan recovery never ran on a headless cold launch** — confirmed. A process kill + iOS BLE-restore relaunch takes `_init()`'s `_background` branch, not `openSession()`, so checkpointed steps sat in prefs forever — the exact scenario the checkpoint exists for. Recovery now runs there too.
- **Recover before enabling live streams** — reordered: recovery + `_resetLivePedometer()` now run *before* `enableLiveStreams()`, closing the window where a frame ingested during the awaited recovery was wiped by the reset behind it.
- **Alarm retry hardening** — added the `_disposed` guard after the await (it was creating a timer `dispose()` could no longer cancel), a staleness check so a retry in flight can't clobber a newly-armed alarm's confirmation state, gave the retry back when the write itself never landed, and extracted the duplicated grace-timer wiring into `_armAlarmGraceTimer`.
- **TZ hold defeated when all pending days are held** — confirmed: `pending` emptied and fell straight through to `latest-finalized-check`, which re-derives `rawDays.last` — one of the days just held. Returns an empty `tz-travel-hold` scope now.
- **Today's exclusion was too broad** (CodeRabbit, Major) — confirmed, and worse than described. `days` is the single input list for the *whole* bundle, so dropping today also removed it from `readiness_glassbox`, the resting-HR trend-shift CUSUM, CTL/ATL/TSB, sleep debt, `_crossDaySri`, `todayStrain`, and `recent` — whose last row is what `_runNotifications` dates every notification from. Today's row is now **flagged** `unsettled` rather than dropped, and `buildCrossDayBundle` nulls only the illness / anomaly / skin-temp-illness inputs for a flagged day. Three regression tests added in `crossday_pipeline_test.dart`.
- **`dayLabelOf()`** used at both live-coverage write sites (the pre-existing `_finalizeLivePedometer` one too, not just the new code).

**Already fixed on the branch, no action**

- Sticky TZ guard (`487194d`), and `kAlgoVersion` — already bumped to **51**, which covers this round's output changes too.

**Declined**

- **Orphan recovery on `openSession()`'s fast-reclaim path** — deliberately not added. Fast reclaim is a foreground resume of a *live* process with an *ongoing* session; the checkpoint there belongs to the session still accumulating in memory. Recovering it would double-count against the in-flight counters, which is the opposite of the bug. Recovery is correct only where counters start fresh (full connect, background cold launch) — both now covered.
- **PR Agent's "`notifyListeners()` never called on `!ok`"** — incorrect; there's no `return` on that path, so it already fell through. Its own scoring notes the suggested code is identical to the existing code.
- **PR Agent's "TZ guard writes inside an isolate"** — `_deriveScope` runs on the main isolate; `LocalDb` calls there are correct and consistent with the rest of the file.

`flutter analyze` clean, `flutter test` **1058 passing** (was 1055).

abdulsaheel:
Second round of bot findings, checked against current code — three were real and are fixed in cc126ee.

**Fixed**

- **`napSub` never reaches the import path** (CodeRabbit, Major) — real. `_deriveDay` (CSV/WHOOP import) didn't pass `napSub`, so it fell back to the `daySub` default and bisected midnight-straddling naps again — on exactly the path this PR set out to fix. Now passes the same buffered slice `prepareDerivationPayload` uses.
- **Orphan recovery could replay an already-banked bout** (CodeRabbit, Major) — real. `_finalizeLivePedometer` writes `live_coverage` *before* clearing the checkpoint, and a kill in that gap leaves a checkpoint whose steps are already durable; `live_coverage` has no uniqueness on the window (it's a plain `SUM`), so recovery re-added them.
  I did **not** take the proposed fix of clearing the checkpoint before the write — that just trades a double-count for a guaranteed loss, and inverts the durable-first ordering this codebase uses everywhere else (same rule as commit-before-ACK). Instead recovery is now idempotent: `LocalDb.hasLiveCoverageWindow(start, end)` and skip on a hit. Safe ordering kept, no data lost either way.
- **TZ guard baseline reset too early** (PR Agent, Medium) — real. The `force` branch re-baselined the guard at scope-*selection* time, before the restage had actually run, so an interrupted restage dropped the hold anyway. Moved to after the full-history pass completes.

**Declined**

- **PR Agent: "`_disposed` guard only reached on the non-throw path"** — incorrect. `try/catch` doesn't exit the function; both paths fall through to the `if (_disposed) return;` below. Its own scoring notes this.
- **PR Agent: "`_maybeDowngradeLiveForBackground` may reset the pedometer"** — verified it doesn't; it only calls `engine.enableHrOnlyLive()`. No change needed (the suggestion was a verification request, not a diff).

`flutter analyze` clean, `flutter test` **1058 passing**.

abdulsaheel:
Third round — all three CodeRabbit findings on cc126ee were real. Fixed in bd4c59d. PR Agent reported "No code suggestions found" for this commit.

**Fixed**

- **`napSub` never reached the LIVE decoded path** (Major) — the important one. My previous commit only fixed `_deriveDay` (import). `_prepareTargetDay` — used by `run()` / `runDays()` / `rescanRecent()`, i.e. **every non-imported day** — still called `toPreparedDay(daySub:, sleepSub:)` with no `napSub`, so it fell back to `daySub` and kept bisecting cross-midnight naps. In other words the headline fix of this PR was inert on the real path. It now loads `[dayStart, dayEnd + napBoundaryBufferSec)` in **one** substrate pass and slices the calendar day back out of it — each `_loadSubstrateRange` spawns its own isolate, so a second load would have doubled the per-day prepare cost; the widened range is ~+12% rows and stays far under `_maxDayRawRows`/`_maxDayRawPages`.
- **TZ hold cleared after skipped/failed days** (Major) — real. `processDay` catches per-day errors and marks the day skipped, so a restage could "finish" with days unresolved and still drop the hold. Now gated on every targeted day actually deriving (a deliberately-kept pruned override day is not counted as a failure).
- **Recovery check-then-insert not atomic** (Major) — real in principle now that two entry points can call it. Made `_recoverOrphanedLiveSession` single-flight, which closes the interleaving window at the source. I did **not** add a unique window key + dedup migration: `live_coverage` may already hold legitimate duplicate-looking rows on existing installs, so creating a `UNIQUE` index inside `openDatabase` risks throwing on the iOS CPU watchdog path for a race that only one process can now reach.

**Partially declined**

- **"Apply the same rule to `runDays(force: true)`"** — `runDays` is the *selected-days* path, not the full-restage entry point; "Re-analyze data" (`reanalyzeAll`) calls `run(profile, heavy: true, force: true)`, which is where the reset lives. Clearing the hold from an arbitrary selected-day subset would be the same premature-clear bug in a different place.

`flutter analyze` clean, `flutter test` **1058 passing**.

abdulsaheel:
Fourth round on bd4c59d — two real, two declined. Fixed in 43d7613.

**Fixed**

- **CodeRabbit: reset the travel guard after a successful *selected* full restage** (Major) — fair pushback on my partial decline last round, and the refined condition resolves my objection. `reanalyzeDays()` (Advanced → select days) goes through `runDays(force: true)`, so a user who selects everything could complete a restage and stay blocked by `tz-travel-hold` forever. Now cleared from that path too — but only when the selection actually covers the whole raw history **and** every target resolved. A partial selection still leaves the hold up, since those days say nothing about the ones being held.
- **PR Agent: durable-first ordering in orphan recovery** (7/10) — correct, and it's a regression I introduced. Removing the checkpoint before writing coverage meant a kill in that gap *lost* the bout — exactly what the checkpoint exists to prevent, and backwards from the commit-before-ACK rule used everywhere else here. Flipped to write-then-drop. This is only safe because of the two things added in the last two commits: the coverage-window check makes replay idempotent, and the method is single-flight. A checkpoint that can never be recovered (malformed JSON, or a window the sanitizer rejects) is still dropped immediately so it can't be retried on every connect forever.

**Declined**

- **PR Agent: "`_disposed` guard missing on the retry-failure path"** — factually wrong; `if (_disposed) return;` sits above `notifyListeners()` and is reached on *both* the throw and the `!rearmed` paths. Its own proposed diff is a no-op that only deletes comments. This is the third variant of this same claim across rounds.
- **PR Agent: "delete `metric_series` rows when a day re-derives as partial"** (3/10) — declined. The suggestion argues against itself mid-paragraph, and its own scoring says so. Deleting the last known-good values on a partial re-derive would make the baseline read a gap instead of the last good day — that is precisely the blank-readiness bug class fixed in edge#108, and `upsertDayResult` skipping the write is what keeps a timed-out second-half pass from overwriting good values in the first place.

`flutter analyze` clean, `flutter test` **1058 passing**.

abdulsaheel:
Fifth round on 43d7613: CodeRabbit posted no new findings (status: "Review completed"). PR Agent raised one, which I'm declining as factually incorrect.

**Declined — "Non-idempotent timezone guard baseline update"**

The claim is that "if a timezone jump happens between two normal derive runs, the second normal run will see `lastOffset == nowOffset` (because the first run already updated it) and return false."

That is not what the code does. `_timezoneTravelSuspected` only writes the baseline on the **non-jump** path:

```dart
final jumped = (nowOffsetMin - lastOffsetMin).abs() >= _tzJumpThresholdMin;
if (!jumped) {
  await LocalDb.putBaseline('tz_travel_guard', ...);
}
return jumped;
```

When a jump *is* detected the stored offset is deliberately left alone, so every subsequent pass compares against the same pre-travel baseline and keeps returning true until a full restage clears it. That was exactly the fix applied in 487194d, and the suggestion would re-open the hole it closed — with the non-jump write removed, an ordinary DST shift (60 min, well under the 180 min threshold) would leave the baseline permanently stale, and a later legitimate 2 h move measured against a months-old offset would read as travel and hold days for no reason.

The one genuine residual is sub-threshold staircase drift (two separate <3 h offset changes summing to more than 3 h are each absorbed). The proposed change doesn't address that either, and a single discrete offset change is what actual travel looks like. Leaving as-is.

Merging — CI green, `flutter analyze` clean, `flutter test` 1058 passing, `kAlgoVersion` at 51.

PR #174, opened 2026-08-02
fix: sleep profile folds every derive pass instead of once per night

abdulsaheel:
### **User description**
A real user export had `"nights": 1348` in `sleep_user_profile`. They have 12 days of data.

The EWMA fold runs on every staging pass for a day — algo bumps, BLE drain re-derives, backfill sweeps — so the same handful of nights folded hundreds of times. Two things break:

- `personalWeight` saturates at 28 nights, so it's pinned at the 0.5 cap from the first sweep
- the EWMA (alpha ~2/15) collapses onto whichever day was re-derived last, not a representative fortnight

Replayed that profile over the same 11 nights. Worst night: **wake 4.3% -> 36.4%, deep 1.9% -> 0.0%**. So the personalization layer was re-creating the exact wake over-call cardioStager exists to avoid. Same bug class as #108, different table.

## fix

- fold at most once per `day_id`, tracked in the profile payload (no migration — `baselines` is free-form JSON)
- don't apply the profile at all until 3 nights. van der Aar 2025: gains need >=3 nights and ~17.5% of subjects get *worse* from personalization, so a 1-2 night profile is pure downside
- discard pre-tracking profiles. Can't repair them — there's no record of which days went in — so they rebuild from cold start, which is the per-night-local path the stager was validated on anyway

## gotcha

Gating the profile to null below the floor would make every fold restart from empty and pin `nights` at 1 forever. The fold uses the loaded profile regardless; only *staging* is withheld. Nearly shipped that.

Logic went into `SleepProfilePolicy` (pure, 15 tests) instead of inline in the engine. One test simulates 12 days x 100 re-derives and asserts exactly 12 folds.

kAlgoVersion 51 -> 52 so days re-stage without the corrupt blend.

1073 tests green, analyze clean.

Separate from OpenStrap/analytics#34 which fixes the staging rules themselves — that one'll need another bump when we repin.

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Fix sleep profile EWMA folding hundreds of times per night instead of once
  - Real user had `nights: 1348` against only 12 days of data
  - Worst-case effect: wake 4.3%→36.4%, deep 1.9%→0.0% on a single night

- Add per-`day_id` fold tracking in profile payload; discard legacy profiles lacking it

- Gate profile influence on staging until ≥3 nights (van der Aar 2025 warm-up floor)

- `kAlgoVersion` bumped 51→52 so all days re-stage without the corrupt blend

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["BLE drain / algo bump\n/ backfill re-derive"] -- "every pass" --> B["_loadSleepUserProfileJson()"]
  B -- "usableProfileJson()\ndiscards legacy" --> C["SleepProfilePolicy"]
  C -- "foldedDays()" --> D["Set of already-folded day_ids"]
  D -- "shouldFold()" --> E{"day_id\nalready folded?"}
  E -- "yes → skip" --> F["No EWMA step\n(idempotent)"]
  E -- "no → fold once" --> G["base.fold(obs)\n+ withFoldedDays()"]
  G --> H["Persist updated profile\nwith folded_days stamp"]
  C -- "shouldBlend(nights)" --> I{"nights >= 3?"}
  I -- "no" --> J["cardioUserProfile = null\n(cold-start staging)"]
  I -- "yes" --> K["cardioUserProfile = loaded\n(personalized staging)"]
```

<tr>
  <td>
    <details>
      <summary><strong>sleep_profile_policy.dart</strong><dd><code>New pure SleepProfilePolicy enforcing fold idempotency and warm-up </code><br><code>gate</code></dd></summary>
<hr>

lib/compute/sleep_profile_policy.dart

<ul><li>New pure-policy class <code>SleepProfilePolicy</code> with no Flutter/DB <br>dependencies<br> <li> <code>shouldFold()</code> gates folding to once per <code>day_id</code> using a <code>folded_days</code> set <br>in the payload<br> <li> <code>shouldBlend()</code> withholds the profile from staging until ≥3 nights are <br>accumulated<br> <li> <code>usableProfileJson()</code> / <code>isLegacy()</code> discard pre-tracking profiles that <br>lack <code>folded_days</code><br> <li> <code>appendFoldedDay()</code> / <code>withFoldedDays()</code> maintain the sorted, capped <br>(400-entry) fold ledger</ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/174/files#diff-5758c8949861f49379ab770e745aa5104f643145cfba9b37fdc54c19740ba550">+130/-0</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>derivation_engine.dart</strong><dd><code>Wire SleepProfilePolicy into derivation engine; bump kAlgoVersion to </code><br><code>52</code></dd></summary>
<hr>

<ul><li><code>kAlgoVersion</code> bumped from 51 to 52 with detailed changelog entry<br> <li> <code>_loadSleepUserProfileJson()</code> now delegates to <br><code>SleepProfilePolicy.usableProfileJson()</code>, discarding legacy profiles<br> <li> Staging worker now separates <code>loaded</code> profile (used for folding) from <br><code>cardioUserProfile</code> (gated by <code>shouldBlend</code>)<br> <li> Fold is now guarded by <code>mayFold</code> (<code>SleepProfilePolicy.shouldFold()</code>); <br>folded result stamped with <code>withFoldedDays()</code></ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/174/files#diff-0e401d8cdb1091d9a9591e9d92c4c5e979293f8480da203b22011a7cf9072ea6">+63/-11</a>&nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Tests</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>sleep_profile_policy_test.dart</strong><dd><code>15 unit tests for SleepProfilePolicy covering all policy branches</code></dd></summary>
<hr>

test/sleep_profile_policy_test.dart

<ul><li>15 new unit tests covering fold idempotency, minimum-nights gate, <br>legacy-payload discard, and bookkeeping<br> <li> Key regression test simulates 12 days × 100 re-derives and asserts <br>exactly 12 folds occur<br> <li> Tests cover cap eviction, JSON round-trip, override suppression, and <br>null/corrupt payload handling</ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/174/files#diff-4a206e94e54033434f545750357989a07f117d624f3d4ee33c0bbfe04c066616">+164/-0</a>&nbsp; </td>

</tr>

</details>

___

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate daily sleep-profile processing.
  * Sleep profiles now build data during the first three nights before personalization begins.
  * Added safeguards for legacy, malformed, overridden, and previously processed sleep data.
  * Improved reliability when simultaneous updates occur, preserving profile changes and recovering safely from errors.
* **Tests**
  * Added comprehensive coverage for profile folding, validation, concurrent updates, data preservation, and recovery scenarios.

### DIFF
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -1264,7 +1264,21 @@ class DerivationEngine {
         payloadJson: candidateJson,
       );
       if (observationJson != null) {
-        await _foldObservationIntoProfile(dayId, observationJson);
+        // BEST-EFFORT, and deliberately isolated from the day's success path.
+        // The fold is bookkeeping; the day's real result is already persisted
+        // above. `updateBaseline` takes an exclusive SQLite write lock, and the
+        // whole point of this change is that two derivation isolates contend
+        // for it — so SQLITE_BUSY here is an EXPECTED outcome, not an
+        // exceptional one. Letting it escape would hit processDay's broad
+        // catch, which calls `_markDaySkipped` and increments `failures`,
+        // throwing away a fully computed day (and holding the timezone) over a
+        // bookkeeping write. Nothing is lost by swallowing it: the day_id never
+        // reaches `folded_days`, so the next pass simply folds it again.
+        try {
+          await _foldObservationIntoProfile(dayId, observationJson);
+        } catch (e) {
+          _log('sleep profile fold skipped for $dayId (day result kept): $e');
+        }
       }
     }
     return candidate;
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -1186,7 +1186,7 @@ class DerivationEngine {
     // with no timeout at all, so a hung staging pass never completed its future
     // — `_running` stayed true and `DeriveScheduler._drain` never returned, i.e.
     // all derivation was dead until app restart.
-    final (candidateJson, updatedProfileJson) =
+    final (candidateJson, observationJson) =
         await _runIsolateCancellable(() {
       try {
         final p = profileJson == null
@@ -1216,7 +1216,7 @@ class DerivationEngine {
       // Fold the MAIN sleep (most epochs) of a freshly-staged night into the
       // rolling profile — done here in the worker because the observations live
       // in THIS isolate's globals. Skipped for overrides. EWMA self-seeds.
-      String? foldedJson;
+      String? observationJson;
       // IDEMPOTENT PER DAY. `fold()` is an EWMA step that also increments
       // `nights`, and this path runs on EVERY staging pass for a day — an
       // algo-version bump, a BLE-drain re-derive, a backfill sweep. Without a
@@ -1237,7 +1237,7 @@ class DerivationEngine {
             // would bake in the profile this worker read before staging began,
             // and a concurrent day may have written a newer one since. The
             // fold happens on the main isolate under the profile lock.
-            foldedJson = jsonEncode({
+            observationJson = jsonEncode({
               'epochs': main.epochs,
               'hr_floor_p5': main.hrFloorP5,
               'hr_floor_p25': main.hrFloorP25,
@@ -1253,7 +1253,7 @@ class DerivationEngine {
           }
         }
       }
-      return (jsonEncode(candidate.toJson()), foldedJson);
+      return (jsonEncode(candidate.toJson()), observationJson);
     }, _perDayTimeout, label: 'sleep-staging $dayId');
     final candidate = SleepSessionCandidate.fromJson(
         (jsonDecode(candidateJson) as Map).cast<String, dynamic>());
@@ -1263,8 +1263,8 @@ class DerivationEngine {
         algoVersion: kAlgoVersion,
         payloadJson: candidateJson,
       );
-      if (updatedProfileJson != null) {
-        await _foldObservationIntoProfile(dayId, updatedProfileJson);
+      if (observationJson != null) {
+        await _foldObservationIntoProfile(dayId, observationJson);
       }
     }
     return candidate;
@@ -1280,42 +1280,55 @@ class DerivationEngine {
   /// a lost day_id, and the day then re-folds forever.
   Future<void> _foldObservationIntoProfile(
       String dayId, String observationJson) async {
-    await SleepProfilePolicy.withProfileLock(() async {
-      final fresh = await _loadSleepUserProfileJson();
-      final freshDays = SleepProfilePolicy.foldedDays(fresh);
+    final Map<String, dynamic> o;
+    try {
+      o = (jsonDecode(observationJson) as Map).cast<String, dynamic>();
+    } catch (_) {
+      return;
+    }
+    double? d(String k) => (o[k] as num?)?.toDouble();
+    final observed = ana.SleepNightObservation(
+      epochs: (o['epochs'] as num?)?.toInt() ?? 0,
+      hrFloorP5: d('hr_floor_p5'),
+      hrFloorP25: d('hr_floor_p25'),
+      hrSleepMedian: d('hr_sleep_median'),
+      hrArousal: d('hr_arousal'),
+      rmssdMed: d('rmssd_med'),
+      rmssdMad: d('rmssd_mad'),
+      enmoStillCut: d('enmo_still_cut'),
+      enmoMoveCut: d('enmo_move_cut'),
+      lfhfMed: d('lfhf_med'),
+      rkMed: d('rk_med'),
+    );
+    // The whole read-modify-write happens inside ONE exclusive DB transaction.
+    // A Dart mutex cannot do this job: `derivationDispatcher` is a
+    // vm:entry-point WorkManager entry that builds its own DerivationEngine in
+    // a SEPARATE background isolate, and a `static` lock has one copy per
+    // isolate — so a background heavy pass and a foreground sweep would each
+    // read the same profile, fold, and clobber the other, losing both the fold
+    // and its day_id from folded_days. SQLite's write lock is cross-connection
+    // and therefore cross-isolate.
+    await LocalDb.updateBaseline('sleep_user_profile', (current) {
+      // Re-derive freshness INSIDE the transaction: the value this day read
+      // before staging is stale by definition, another lane may have folded
+      // since. Returning null leaves the row untouched.
+      final usable = SleepProfilePolicy.usableProfileJson(current);
+      final freshDays = SleepProfilePolicy.foldedDays(usable);
       if (!SleepProfilePolicy.shouldFold(
           alreadyFolded: freshDays, dayId: dayId, hasOverride: false)) {
-        return; // another lane folded this day while we were staging
+        return null;
       }
       final ana.SleepUserProfile base;
       try {
-        base = fresh == null
+        base = usable == null
             ? const ana.SleepUserProfile()
             : ana.SleepUserProfile.fromJson(
-                (jsonDecode(fresh) as Map).cast<String, dynamic>());
+                (jsonDecode(usable) as Map).cast<String, dynamic>());
       } catch (_) {
-        return; // unreadable profile — leave it for the cold-start path
+        return null; // unreadable — leave it for the cold-start path
       }
-      final o = (jsonDecode(observationJson) as Map).cast<String, dynamic>();
-      double? d(String k) => (o[k] as num?)?.toDouble();
-      final folded = base.fold(ana.SleepNightObservation(
-        epochs: (o['epochs'] as num?)?.toInt() ?? 0,
-        hrFloorP5: d('hr_floor_p5'),
-        hrFloorP25: d('hr_floor_p25'),
-        hrSleepMedian: d('hr_sleep_median'),
-        hrArousal: d('hr_arousal'),
-        rmssdMed: d('rmssd_med'),
-        rmssdMad: d('rmssd_mad'),
-        enmoStillCut: d('enmo_still_cut'),
-        enmoMoveCut: d('enmo_move_cut'),
-        lfhfMed: d('lfhf_med'),
-        rkMed: d('rk_med'),
-      ));
-      await LocalDb.putBaseline(
-        'sleep_user_profile',
-        jsonEncode(SleepProfilePolicy.withFoldedDays(
-            folded.toJson(), freshDays, dayId)),
-      );
+      return jsonEncode(SleepProfilePolicy.withFoldedDays(
+          base.fold(observed).toJson(), freshDays, dayId));
     });
   }
 
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -1272,12 +1272,28 @@ class DerivationEngine {
         // exceptional one. Letting it escape would hit processDay's broad
         // catch, which calls `_markDaySkipped` and increments `failures`,
         // throwing away a fully computed day (and holding the timezone) over a
-        // bookkeeping write. Nothing is lost by swallowing it: the day_id never
-        // reaches `folded_days`, so the next pass simply folds it again.
+        // bookkeeping write.
+        //
+        // KNOWN LIMITATION — a swallowed failure here is PERMANENT for this
+        // day, not retried. Once the day finalizes, the cached-candidate
+        // short-circuit at the top of this method returns before staging runs,
+        // so `observationJson` is never regenerated and the fold never happens.
+        // Same for a day whose override is later removed if it already has a
+        // cached candidate from before the override.
+        //
+        // Accepted deliberately rather than fixed: the profile is an EWMA with
+        // a ~14-night horizon and a hard 0.5 blend cap, so one missing night is
+        // a small perturbation, whereas a retry path needs durable pending
+        // state and a way to distinguish "failed, retry" from "declined
+        // permanently" (a <120-epoch nap never folds, and would otherwise
+        // bypass the candidate cache and re-stage on every sweep forever).
+        // If the fold ever stops being best-effort, that state machine is the
+        // thing to build — do not simply bypass the cache.
         try {
           await _foldObservationIntoProfile(dayId, observationJson);
         } catch (e) {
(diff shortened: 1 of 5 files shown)


abdulsaheel:
Checked the one finding. It's real — fixed.

## the race

Verified the path is genuinely concurrent: `processDay` is run through `runWithConcurrency(orderedDays, _deriveConcurrency, ...)` and `DerivePacing.concurrency` returns `min(cores, maxForegroundConcurrency)` in the foreground, so >1 on any real device.

So: two days read the same `sleep_user_profile`, both await their staging isolate, both decide to fold, later write clobbers the earlier one. Loses the fold **and** its `day_id` from `folded_days` — so that day re-folds on the next sweep and `nights` drifts up again. Which is the exact accounting corruption this PR is fixing, just at a slower rate. The idempotence tests didn't catch it because they test the policy in isolation, as you said.

## fix

The worker now returns the raw **observation** instead of a pre-folded profile — folding in the isolate would bake in whatever profile it read before staging started, which is stale by definition. The fold moved to the main isolate inside `SleepProfilePolicy.withProfileLock`, which re-reads the profile and re-checks `shouldFold` inside the critical section.

Went with the atomic read-modify-write rather than accumulate-and-merge-after-the-run: the lock is only held across the DB read/write, never across the isolate, so day-level concurrency is untouched, and it doesn't require restructuring the sweep.

3 tests, including one that models N days folding concurrently and asserts every `day_id` survives, plus one that a throwing action releases the lock rather than deadlocking the rest of the sweep.

18 policy tests, 1073 total, analyze clean.

abdulsaheel:
Both right. Second one especially — thanks for catching it.

## the lock doesn't cross isolates

Verified before changing anything: `derivationDispatcher` is a `@pragma('vm:entry-point')` WorkManager entry that constructs its **own** `DerivationEngine(background: true)` and calls `engine.run(...)`. That's a separate background isolate, and a Dart `static` has one copy per isolate. So a background heavy pass and a foreground sweep would each hold "the" lock and clobber each other anyway.

Worse than nothing, really — it looked sufficient. Removed it and left a comment saying why, so nobody re-adds one.

Replaced with `LocalDb.updateBaseline`: the whole read-modify-write happens inside one exclusive SQLite transaction. `exclusive: true` issues BEGIN IMMEDIATE so the write lock is taken up front rather than failing to upgrade under WAL when another connection holds it. SQLite's lock is cross-connection, so it holds across isolates and processes — which is the guarantee that was actually needed.

## the vacuous test

You're right and it's a bit embarrassing. Seeding `'2026-07-30'` into `folded_days` meant both lanes skipped on their first read, so `folds == 0` regardless. Deleting the lock left it green. It asserted nothing.

Rewrote that group to model the transaction body at commit time, which is what the exclusive write lock actually guarantees:

- two lanes folding the **same** day from an **empty** profile commit exactly once
- distinct days all survive, none lost
- a stale pre-staging read cannot resurrect an already-folded day (asserts the stale view *would* wrongly permit it, and that deciding against the committed row declines) — that's the specific thing the re-check inside the transaction buys
- a legacy row rebuilds from 0, not from 1348

## rename

Done — `foldedJson` / `updatedProfileJson` → `observationJson`. Fair point that a name saying "profile" in the one path built to avoid writing a stale profile is asking for trouble.

## PR Agent's "lock not reset on throw"

Moot now, the lock is gone. For the record it did release correctly (`.whenComplete`), and there was a test for it — but the whole mechanism was the wrong one.

1077 tests, analyze clean.

abdulsaheel:
Both valid. The first one is a regression I introduced in the previous round, so thanks for staying on it.

## fold failure was failing the whole day

Verified the path: `processDay` has a broad `catch (e)` at both call sites that calls `_markDaySkipped` and increments `failures`, which holds the timezone. And `_markDaySkipped` is the same path that was a P0 for destroying good `day_result` rows.

What makes it worse is that I caused it. Round 2 introduced an exclusive SQLite write lock **specifically so two derivation isolates contend for it** — so `SQLITE_BUSY` is an expected outcome here, not an exceptional one. I raised the probability of an exception while it was wired to a destructive handler. A fully computed, already-persisted day would get discarded because a bookkeeping write lost a lock race.

Wrapped in try/catch with a log line. Nothing is lost: the `day_id` never reaches `folded_days`, so the next pass folds it again.

## direct tests for updateBaseline

Fair — it's the new cross-isolate primitive and had no coverage of its own. Added `test/db_update_baseline_test.dart` on the existing `sqflite_common_ffi` harness, covering the three you listed plus a few:

- `transform` sees `null` for an absent key, and can create the row
- a `null` return leaves the row byte-identical, `updated_at` included
- a non-null return replaces the payload and advances `updated_at`
- sequential accumulate — every update observes the previous commit
- concurrent accumulate, and concurrent set-union (the real `folded_days` shape)
- a throwing transform rolls back and doesn't wedge the connection

**And I checked they actually bite**, since last round I shipped a test that asserted nothing. Temporarily swapped `updateBaseline` for a naive read-then-write:

```
concurrent accumulate    Expected: <20>   Actual: <1>
concurrent set-union     Expected: 20 days   Actual: ['2026-07-29']
```

19 of 20 increments and 19 of 20 day_ids lost. Restored, all green.

1084 tests, analyze clean.

abdulsaheel:
PR Agent's persistent review flagged three things. Checked all three — none is a bug, but two pointed at untested corners so they get tests rather than a "works on my machine".

## "Syntax error in test helper" — incorrect

The claim is that `SleepProfilePolicy.foldedDaysKey: ?foldedDays` is invalid Dart, so the file won't compile and none of the policy tests run.

`key: ?value` is **null-aware element** syntax. Valid on our SDK (3.11.4), and the analyzer's own `use_null_aware_elements` lint is literally what asked me to write it that way — I had it as a conditional `if` first and the analyzer flagged that.

Evidence:

```
$ flutter test test/sleep_profile_policy_test.dart
00:00 +19: All tests passed!

$ flutter analyze test/sleep_profile_policy_test.dart
No issues found!
```

Also "no CI gate runs on PRs" is wrong — `test.yml` triggers on `pull_request`, and `gh pr checks 174` shows `test  pass  3m33s` on this PR.

## "Legacy discard race" — not reachable, now covered

The scenario: several lanes all observe the pre-tracking row simultaneously, each treats it as a cold start, each writes a profile containing only its own day, so the rebuild drops folds.

It can't happen through `updateBaseline`. `BEGIN IMMEDIATE` serialises the transactions, so only the **first** lane ever sees the legacy row; by the time the second runs, the row already carries `folded_days` and is no longer legacy. The read is inside the same transaction as the write.

Rather than argue it, added `concurrent LEGACY discard: the rebuild loses nothing either` — 15 concurrent lanes over a `nights: 1348` row, asserting `nights == 15`, every `day_id` present, and the 1348 gone.

## "Observation silently dropped on override path" — correct by design, now pinned

The write-up says "this is correct behavior" three times and is partly self-contradictory (says override days are "permanently excluded", then correctly notes a later non-override re-derive would fold them). The second reading is right: skipping an override records nothing in `folded_days`, so the day stays eligible and the behaviour is self-healing.

But "not tested" was fair. Added `skipping an override does NOT blacklist the day forever` — skip under override records nothing, removing the override makes it foldable, and it still folds only once after that. Worth pinning because the tempting alternative (marking it folded to "remember we skipped it") would exclude that night from the profile permanently.

1086 tests, analyze clean.

Note: CodeRabbit shows `Review rate limited` on the last two commits, so it hasn't re-reviewed 45f0939 or 365c6f8 yet.

abdulsaheel:
@coderabbitai review

abdulsaheel:
Both correct. Neither gets a code fix from me, and I want to be explicit about why rather than quietly closing them.

## the fold-retry gap — you're right, and my comment was wrong

Verified: `_sleepCandidateForDay` returns the cached finalized candidate at Lines 1133-1150 before staging runs, so `observationJson` is never regenerated. My round-3 comment claimed a swallowed failure was harmless because "the next pass simply folds it again". **That is false.** A swallowed fold failure is permanent for that day, and the same path blocks re-folding after an override is removed if the day already had a candidate cached.

I have **not** built the retry. Reasoning, so it can be argued with:

- the profile is an EWMA over a ~14-night horizon with a hard 0.5 blend cap, so one missing night is a small perturbation — this is genuinely best-effort bookkeeping
- the naive version of your second suggestion (bypass the cache until the day is in `folded_days`) is actively wrong: a <120-epoch nap never folds, so it would never enter the set, so it would bypass the cache and re-stage **every sweep forever**. Staging is the expensive part.
- doing it properly needs durable pending-observation state plus a distinction between "failed, retry" and "declined permanently". That's a state machine, and this PR has already had two rounds where my fix for one finding created the next one. I'd rather not add one more under momentum.

So: documented as a KNOWN LIMITATION at the call site, with the reasoning and an explicit "do not simply bypass the cache" note for whoever picks it up. Happy to build the retry if you'd rather — say the word.

## the override test was overclaiming

Related, and my fault. `skipping an override does NOT blacklist the day forever` reads as an end-to-end guarantee. It only asserts the **policy** keeps returning true — the engine often won't re-fold, for exactly the reason above. Renamed to `skipping an override records nothing, so the POLICY stays eligible` with a scope note pointing at the limitation.

That's twice now in this PR that I've shipped a test whose name promised more than it checked.

## cross-isolate coverage

Fair. The concurrent tests run in one isolate on one `sqflite_ffi` connection, so they prove atomicity under interleaved async access — they do fail 19/20 against a naive read-then-write — but not two OS-level connections contending, which is the actual scenario.

I didn't add the spawned-isolate test: `LocalDb`'s singleton/static setup has no entry point for opening the same file from a second isolate, so it'd need a test-only seam, and I'm wary of adding production surface purely for a test at this point in the PR. Instead the file header now states plainly what the tests do and don't establish, and that the cross-isolate guarantee rests on SQLite's documented `BEGIN IMMEDIATE` locking rather than on these tests. Flagged as worth building if this primitive picks up more callers.

1086 tests, analyze clean.

abdulsaheel:
Three focus areas in the updated persistent review. One acted on, two need no change.

## "Cap evicts by sort order, not insertion age" — fair, asserted

Narrow but correct. `appendFoldedDay` evicts by **lexicographic** order, which only equals chronological order for zero-padded ISO dates. A UUID or epoch string would break the sort, so a recent day could be evicted while an older one is kept — and an evicted day passes `shouldFold` again, which is precisely the double-fold this class exists to prevent. And it'd be invisible in the payload.

Only `day_label.dart` values reach this today, so it's a precondition, not a bug. Added the assert and documented why the sort is load-bearing.

Nice side effect: the assert immediately failed **my own** cap test, which was generating fake labels like `2020-01-00000`. Rewrote it to walk real dates and added coverage for eviction across a month and year boundary, plus the non-date rejection.

## "Fold skipped permanently on SQLITE_BUSY" — already documented

Same gap CodeRabbit raised in the previous round. Confirmed real, deliberately not fixed, documented as a KNOWN LIMITATION at the call site with the reasoning (EWMA over ~14 nights at a 0.5 cap, and the naive fix — bypassing the candidate cache — would make a <120-epoch nap re-stage every sweep forever). Details in my earlier reply.

## "Stale mayFold pre-check" — not reachable

The substantive half of this was: *"if that drain has side effects (clearing the accumulator), a subsequent legitimate fold attempt for a different day in the same isolate lifetime could see an empty observation list. Whether the isolate is reused across days in the same sweep determines whether this matters."*

Checked — it isn't reused. `_runIsolateCancellable` does `Isolate.spawn` per call with `onError`/`onExit` wired, and the isolate exits after sending its result. One fresh isolate per day, so `_cardioObservations` dies with it and can't starve another day.

The remainder is a handful of serialized doubles that the transaction re-check occasionally declines. That's the intended design rather than waste: `mayFold` can go stale during ~90s of staging no matter how fresh the pre-check is, which is exactly why the authoritative check lives inside the transaction.

1088 tests, analyze clean.

abdulsaheel:
Persistent review is down to two focus areas on `7d9c8e4`, both already answered above (the SQLITE_BUSY gap is documented as a known limitation; the stale `mayFold` drain concern isn't reachable because `_runIsolateCancellable` spawns a fresh isolate per day). The cap-eviction one dropped off after the assert landed.

On the current code suggestions — nothing actionable, but for the record:

**"Warm-up gate driven by tracked distinct nights, not EWMA counter"** (importance 3). The stated divergence doesn't occur: a skipped fold increments *neither* `nights` nor `folded_days`, so they stay in step. They only part company past the 400-entry cap, where `folded_days.length` plateaus while `nights` keeps climbing — and both are far above the 3-night gate by then, so the gate behaves identically. The proposed diff also just adds an unused method beside `shouldBlend` without changing it, which PR Agent's own rationale notes.

**"Ensure exclusive write lock is released on transform exception"**. The suggested diff is a semantic no-op (`try { next = transform(current); } catch (_) { rethrow; }` is exactly what already happens implicitly). The real question inside it — *"confirm sqflite releases the write lock on exception, as a stuck lock would wedge all subsequent baseline writes"* — is worth having pinned, and it already is: `a throwing transform rolls back and leaves the row intact` throws inside the transaction, asserts the row is unchanged, **then performs another successful `updateBaseline`**. A held lock would hang that second call.

**"Fix invalid null-aware map value syntax"** (importance 7). Repeat of the earlier claim. Still incorrect — `key: ?value` is null-aware element syntax, valid on SDK 3.11.4, and `use_null_aware_elements` is the lint that asked for it. Evidence unchanged: `flutter test` on that file reports `+19: All tests passed`, `flutter analyze` reports `No issues found!`, and the `test` job passes on this PR. Not changing working code on a claim that three independent checks contradict.

**"Remove redundant expect that fires before skip takes effect"** — this one targets `test/derivation_pipeline_test.dart` in **PR #154**, not this PR. That file isn't in this diff (`derivation_engine.dart`, `sleep_profile_policy.dart`, `db.dart`, and two test files are). Stale entry accumulated in the suggestions comment.

1088 tests, analyze clean. Nothing further from me unless a human reviewer wants the fold-retry state machine built.

Issue #160, opened 2026-07-26
Encoding error when important data from NOOP


Amau300305:
### What happened?

When uploading my NOOP data to edge it says: NOOP failed: FormatException: Invalid UTF-8 byte (at offset 10)

### App version

Latest

### Platform

iOS (TestFlight)

### Phone model and OS version

iPhone 15 pro Max 

### Which band?

WHOOP 4.0

### Screenshots

<img width="645" height="1398" alt="Image" src="https://github.com/user-attachments/assets/0f22b6f4-74ec-4d6f-943c-a95471248048" />

### Quick checks

- [x] The official WHOOP app is not also connected to this band
- [x] I've searched existing issues for this

abdulsaheel:
Hi, which version of Noop were you using? Could you also share the source link?

Amau300305:
I am on the 9.1.0 by ryanbr on GitHub. You want the sourcelink of the app or my data?

abdulsaheel:
of the app, github link basically

Amau300305:
https://github.com/ryanbr/noop/releases/tag/v9.1.0

MrUphill:
Hi @abdulsaheel 
I can reproduce this issue. It seems the NOOP backup file is a compressed file with a sqlite file. Edge is expecting a CSV File?

Jenssaibe:
Same here. 
Noop 9.2.1

abdulsaheel:
Hi, I hope you guys are exporting from this screen:

<img width="2160" height="2880" alt="Image" src="https://github.com/user-attachments/assets/78314f97-14f2-4b39-a852-059541c99e51" />

If yes, please share the same file with me via Discord DM. If not, please export it from this screen. If you still encounter the same issue, send me the database file via Discord DM.

Amau300305:
I cannot load my data on both solutions

Jenssaibe:
It is only the last 24hours, is there any way to import more data from Noop ?

abdulsaheel:
can someone please share the file, thanks.

Amau300305:
[noop-raw-sensors-1785567300.csv](https://github.com/user-attachments/files/30612829/noop-raw-sensors-1785567300.csv)

Jenssaibe:
> can someone please share the file, thanks.
> 

Just to be sure, you want a full Noop backup (*.noopbak) ?

abdulsaheel:
> > can someone please share the file, thanks.
> 
[...]

PR #176, opened 2026-08-02
NOOP import: recover the band step counter after schema drift (#160)

abdulsaheel:
### **User description**
Investigating #160 turned up **three separate problems**. This PR fixes one of them, and documents the other two rather than silently leaving them implied.

## 1. The reported crash is a `.noopbak`, not a CSV — NOT fixed here

`.noopbak` is a **ZIP** (wrapping SQLite). A ZIP local file header puts the DOS mod-time low byte at **offset 10**, which is ≥`0x80` in most files. Reproduced byte-exactly:

```
50 4b 03 04 14 00 00 00 08 00 fc 8d 02 5d 29 ba
byte@10 = 0xfc → "can't decode byte 0xfc in position 10"
```

That is character-for-character the `FormatException: Invalid UTF-8 byte (at offset 10)` in the issue. `import_screen.dart` picks `FileType.any` and `noop_import.dart` pipes it straight into `utf8.decoder`, so any non-text file becomes a raw decoder exception. @MrUphill diagnosed this correctly in the thread.

**Deliberately out of scope here** — it wants a magic-byte sniff plus a human-readable error (and arguably real `.noopbak` support), which is a separate change from the data fix below.

## 2. The schema drift — what this PR fixes

NOOP's current export (9.1/9.2):

```
unix_s,iso_utc,stream,hr_bpm,rr_ms,grav_x,grav_y,grav_z,step_counter,ppg_bpm,ppg_conf,
spo2_red,spo2_ir,skintemp_raw,resp_raw,band_sleep_state,event_kind,event_payload
```

- `band_sleep_state` was **inserted at index 15**, shifting `event_kind`/`event_payload` to 16/17. Reading by NAME meant **nothing misparsed** — the existing defensive design held.
- New streams: `steps`, `band_sleep_state`, `ppghr`.
- `spo2` and `resp` rows are **no longer emitted at all**, though their columns survive in the header.

The real loss was `steps` landing in the importer's `default:` branch. Every imported day reported **0 steps** while the band's own counter had measured them — 2,572 over the 3.5 h in the export attached to the issue.

### The fix

The cumulative counter is differenced into contiguous runs and banked into **`live_coverage`** — the same table the live 100 Hz pedometer writes. That seam needs **no change to `Substrate` or the derivation engine**: `liveStepsForDay` picks up the real count, and `coverageWindowsOverlapping` keeps the 1 Hz estimate from double-counting those minutes. Imported and live days are counted identically.

`stepRuns` is pure and unit-tested:
- splits on gaps > 60 s, so the export's 20.5 h hole never becomes one window claiming to cover the day
- sums only **positive** deltas — a negative delta is a reboot reset, not −24,000 steps
- **drops deltas across a run boundary** rather than attributing steps to a window we have no samples for
- emits nothing for a 0-step run, which would suppress a real 1 Hz estimate while contributing nothing

`_flushStepCoverage` is **idempotent** via `hasLiveCoverageWindow`: `live_coverage` is an append-only SUM with no uniqueness on the window, so re-importing the same export would otherwise double every step.

Verified against the real file from the issue:

```
STEPS=2572   coverage 1785480900..1785493571
"steps":{"value":2572,"real_100hz":2572,"tier":"HIGH"}     ← was steps=0.0
```

2,572 matches the counter range (24302 → 26874) measured independently, and the window covers only the 3.5 h block, not the hole.

### Deliberately not consumed

| stream | measured | why skipped |
|---|---|---|
| `band_sleep_state` | constant `0` across all 12,663 rows | no signal, and a second sleep source would contradict `segmentSleep` |
| `ppghr` | 151 samples, conf 0.28–0.81 | `hr` already covers all 12,692 seconds; PPG-derived HR isn't trusted as a substitute |

Both are documented in the file header with the reasoning, so the next reader doesn't have to re-derive it.

## 3. Why the data still looks thin — no code change

Worth stating plainly for the reporters: the CSV **was importing fine all along** (`days=2 rows=66210 lateRows=0`). It just doesn't contain much:

```
07-31T06  8.3% · T07 100% · T08 100% · T09 99.9% · T10 43.6% · [gap 73672 s = 20.5 h] · 08-01T06 0.8%
```

One 3.5 h daytime block, then a 20.5 h hole. No overnight ⇒ no sleep ⇒ RMSSD/readiness/strain are legitimately null. That matches @Jenssaibe's "it is only the last 24 hours" — a **NOOP export-scope limitation, not an edge bug**. No warning UI added; the import reports honestly and the metrics abstain rather than fabricate.

## Version

`kAlgoVersion` 53 → 54. Only the `steps`/`active_min` block of **imported** days changes; no live-sync output moves.

⚠️ The bump does **not** retro-fix an existing import — imported days are force-finalized snapshots with no stored raw to recompute from, so an already-imported day needs a re-import to pick its steps up. Called out in the changelog comment.

## Tests

`test/noop_schema_drift_test.dart` — 11 tests. Seven pure `stepRuns` cases (counter reset, gap split, cross-boundary non-attribution, zero-step run, order independence) and four end-to-end against the **current** header, including re-import idempotency, an export with no `steps` stream, and an unknown future stream.

There was previously **no test that parsed a real NOOP CSV at all** — only the pure `decideRow` ordering contract. That is precisely why this drift shipped unnoticed, and the end-to-end cases are the part that would have caught it.

Full suite: **1099 passing**, `flutter analyze` clean.

## Note for reviewers

One cosmetic honesty gap left alone: an imported day's steps block reports `inputs_used: ["live_100hz_pedometer", ...]` and the note "real 100 Hz count only". Accurate as *"the real-count channel"*, inaccurate as provenance for a NOOP import. Fixing it means threading a source flag through `_stepsAndEnergy`/`_DeriveInput`, which felt like a poor trade against touching the shared derive path in this PR. Happy to do it if you'd rather.

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- NOOP CSV importer now recovers real step counts from the `step_counter` column instead of silently dropping them (2,572+ steps lost per import in issue #160).

- `stepRuns()` differences the cumulative counter into contiguous windows, splits on gaps >60 s, ignores negative deltas (reboots), and banks results into `live_coverage` — idempotent via `hasLiveCoverageWindow`.

- `kAlgoVersion` bumped from 53 → 54; only imported days' `steps`/`active_min` block changes; no live-sync output moves.

- New regression test file pins the schema-drift fix, `stepRuns` pure logic, idempotent re-import, and graceful handling of unknown future streams.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["NOOP CSV\n(steps stream)"] -- "parse step_counter\nby column NAME" --> B["stepRuns()\npure function"]
  B -- "split on gaps >60s\nsum positive deltas" --> C["StepRun list"]
  C -- "_flushStepCoverage()\nidempotent via hasLiveCoverageWindow" --> D["live_coverage\ntable (LocalDb)"]
  D -- "liveStepsForDay()\ncoverageWindowsOverlapping()" --> E["DerivationEngine\n(kAlgoVersion=54)"]
  E --> F["steps / active_min\nin day_result"]
```

<tr>
  <td>
    <details>
      <summary><strong>derivation_engine.dart</strong><dd><code>Bump kAlgoVersion to 54 for NOOP step recovery</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

<ul><li>Bumps <code>kAlgoVersion</code> from 53 to 54.<br> <li> Adds changelog entry explaining that only imported days' <br><code>steps</code>/<code>active_min</code> block changes; no live-sync output moves.<br> <li> Notes that already-imported days need a re-import to pick up real <br>steps (no stored raw to recompute from).</ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/176/files#diff-0e401d8cdb1091d9a9591e9d92c4c5e979293f8480da203b22011a7cf9072ea6">+14/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>noop_import.dart</strong><dd><code>Recover real step counts from NOOP step_counter stream</code>&nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

lib/import/noop_import.dart

<ul><li>Updates file header to document the current NOOP 9.1/9.2 schema (new <br><code>band_sleep_state</code> column at index 15, new streams <br><code>steps</code>/<code>band_sleep_state</code>/<code>ppghr</code>, absent <code>spo2</code>/<code>resp</code> rows) and explains what <br>is consumed vs. deliberately ignored.<br> <li> Adds <code>StepRun</code> value class and <code>stepRuns()</code> pure static method that <br>differences the cumulative <code>step_counter</code> into contiguous windows, <br>splits on gaps >60 s, ignores negative deltas (counter resets), and <br>emits nothing for zero-step runs.<br> <li> Adds <code>_flushStepCoverage()</code> which banks <code>StepRun</code>s into <code>live_coverage</code> <br>idempotently via <code>hasLiveCoverageWindow</code>, returning the steps actually <br>banked.<br> <li> Wires step accumulation into the main parse loop (per-date <code>stepsByDate</code> <br>map), flushes before each <code>deriveAndPrune</code> call, and handles the EOF and <br>steps-only-date edge cases; adds <code>steps</code> field to <code>NoopImportResult</code>.</ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/176/files#diff-144d9ba6b7f5044334a71db50120c98d507cff16562d93fb1967fc5a994a737f">+150/-9</a>&nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Tests</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>noop_schema_drift_test.dart</strong><dd><code>Add regression tests for NOOP schema drift and step recovery</code></dd></summary>
<hr>

test/noop_schema_drift_test.dart

<ul><li>New regression test file (213 lines) covering the NOOP schema-drift <br>fix.<br> <li> Pure unit tests for <code>stepRuns()</code>: positive deltas, gap splitting, <br>counter reset, cross-boundary attribution, zero-step run, <br>empty/single-sample input, and order-independence.<br> <li> End-to-end integration tests using sqflite_ffi: verifies steps are <br>banked correctly, re-import is idempotent (no double-counting), <br>shifted <code>event_kind</code>/<code>event_payload</code> columns do not misparse, missing <br><code>steps</code> stream imports cleanly with 0 steps, and unknown future streams <br>are skipped without error.</ul>

</details>

  </td>
  <td><a href="https://github.com/OpenStrap/edge/pull/176/files#diff-877f1125a0f686f5b51e883117d19cc40915dab934e64427d4ec1657c875172c">+213/-0</a>&nbsp; </td>

</tr>

</details>

___

## Summary by CodeRabbit

* **New Features**
  * CSV imports now preserve real step counts from supported step-counter data.
  * Imported step data is applied consistently with live activity data without double-counting estimates.
  * Imports now handle current schema variations, quoted events, missing streams, and unknown streams more reliably.

* **Bug Fixes**
  * Corrected overlapping or repeated imports to prevent duplicate step coverage.
  * Improved handling of counter resets, gaps, and incomplete import dates.

### DIFF
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -391,7 +391,20 @@ import 'substrate.dart';
 // hypnogram, stage minutes and sleep-derived scalars change, so every day must
 // re-derive. Also picks up the protocol realtimeRr bound (live HRV no longer
 // sees implausible sub-100ms "beats" from a misaligned 0x28 frame).
-const int kAlgoVersion = 53;
+// v54: NOOP CSV imports now bank the export's `step_counter` as REAL steps.
+// NOOP's schema gained a `steps` stream (and a `band_sleep_state` column that
+// shifted event_kind/event_payload) — the importer read columns by name so it
+// never misparsed, but it dropped `steps` into its default branch and every
+// imported day reported steps = 0 while the band had actually counted them
+// (2,572 over the 3.5 h in the OpenStrap/edge#160 export). The counter is now
+// differenced into contiguous runs and written to `live_coverage`, the same
+// table the live 100 Hz pedometer uses, so imported and live days count steps
+// identically and the 1 Hz estimate still cannot double-count those minutes.
+// Only the `steps`/`active_min` block of IMPORTED days changes; no live-sync
+// output moves. NOTE this bump does not retro-fix an existing import — imported
+// days are force-finalized snapshots with no stored raw to recompute from, so
+// an already-imported day needs a re-import to pick its steps up.
+const int kAlgoVersion = 54;
 
 // Fold idempotency, the minimum-nights warm-up, and legacy-payload handling
 // all live in SleepProfilePolicy (pure, unit-tested) — see
--- a/lib/import/noop_import.dart
+++ b/lib/import/noop_import.dart
@@ -353,27 +353,61 @@ class NoopImporter {
   /// A run with zero steps still yields no window: `live_coverage` exists to
   /// suppress the 1 Hz estimate over minutes the real counter already covered,
   /// and a 0-step run would suppress a real estimate while contributing nothing.
-  static List<StepRun> stepRuns(Map<int, int> samples) {
+  ///
+  /// [covered] lists time spans ALREADY banked in `live_coverage` (device-time
+  /// seconds, inclusive). A per-second delta whose interval intersects one of
+  /// them is skipped and BREAKS the run, so those steps are never banked twice.
+  /// This is what makes a re-import safe in general rather than only for a
+  /// byte-identical file: an exact-window check alone is defeated the moment the
+  /// user exports again over a longer span (09:00-09:20 then 09:00-09:40), where
+  /// the run boundary shifts, no exact window matches, and the overlap is banked
+  /// a second time — measured at 3,598 steps against a true 2,399 before this.
+  /// Clipping is exact, not pro-rated: we hold the counter value at every
+  /// second, so the uncovered sub-intervals are summed from real deltas.
+  /// It also means an imported span cannot double-count against a LIVE 100 Hz
+  /// pedometer window, which shares this table.
+  static List<StepRun> stepRuns(
+    Map<int, int> samples, {
+    List<List<int>> covered = const [],
+  }) {
     if (samples.isEmpty) return const [];
     final ts = samples.keys.toList()..sort();
+
+    // Does the half-open delta interval (a, b] touch anything already banked?
+    bool isCovered(int a, int b) {
+      for (final w in covered) {
+        if (b > w[0] && a < w[1]) return true;
+      }
+      return false;
+    }
+
     final out = <StepRun>[];
-    var runStart = ts.first;
+    int? runStart, runLast;
     var runSteps = 0;
+    void closeRun() {
+      if (runStart != null &&
+          runLast != null &&
+          runSteps > 0 &&
+          runLast! > runStart!) {
+        out.add(StepRun(runStart!, runLast!, runSteps));
+      }
+      runStart = null;
+      runLast = null;
+      runSteps = 0;
+    }
+
     for (var i = 1; i <= ts.length; i++) {
-      final broke = i == ts.length || ts[i] - ts[i - 1] > stepRunMaxGapSec;
-      if (!broke) {
-        final d = samples[ts[i]]! - samples[ts[i - 1]]!;
-        if (d > 0) runSteps += d;
+      final bankable = i < ts.length &&
+          ts[i] - ts[i - 1] <= stepRunMaxGapSec &&
+          !isCovered(ts[i - 1], ts[i]);
+      if (!bankable) {
+        closeRun();
         continue;
       }
-      final runEnd = ts[i - 1];
-      if (runSteps > 0 && runEnd > runStart) {
-        out.add(StepRun(runStart, runEnd, runSteps));
-      }
-      if (i < ts.length) {
-        runStart = ts[i];
-        runSteps = 0;
-      }
+      runStart ??= ts[i - 1];
+      final d = samples[ts[i]]! - samples[ts[i - 1]]!;
+      if (d > 0) runSteps += d;
+      runLast = ts[i];
     }
     return out;
   }
@@ -383,16 +417,23 @@ class NoopImporter {
   /// estimate (`coverageWindowsOverlapping`) — the same contract the live 100 Hz
   /// pedometer uses, so imported and live days are counted identically.
   ///
-  /// IDEMPOTENT: `live_coverage` is an append-only SUM with no uniqueness on the
-  /// window, so re-importing the same export would otherwise double every step.
-  /// Each window is checked with `hasLiveCoverageWindow` first.
+  /// IDEMPOTENT BY TIME SPAN, not by exact window: `live_coverage` is an
+  /// append-only SUM with no uniqueness constraint, so anything already banked
+  /// over a second must not be banked again. The spans covering this batch are
+  /// read back and passed to [stepRuns], which clips them out. An exact-window
+  /// check is deliberately NOT used — it only catches a byte-identical
+  /// re-import and silently double-counts the overlap when the user exports
+  /// again over a longer span (see the note on [stepRuns]).
   ///
-  /// Returns the steps actually banked (0 for windows already present).
+  /// Returns the steps actually banked (0 when the span was fully covered).
   static Future<int> _flushStepCoverage(
       Map<int, int> stepSamples, String date) async {
+    if (stepSamples.isEmpty) return 0;
+    final ts = stepSamples.keys.toList()..sort();
+    final existing =
+        await LocalDb.coverageWindowsOverlapping(ts.first, ts.last + 1);
     var banked = 0;
-    for (final r in stepRuns(stepSamples)) {
-      if (await LocalDb.hasLiveCoverageWindow(r.startSec, r.endSec)) continue;
+    for (final r in stepRuns(stepSamples, covered: existing)) {
       await LocalDb.addLiveCoverage(r.startSec, r.endSec, r.steps, date);
       banked += r.steps;
     }
--- a/lib/import/noop_import.dart
+++ b/lib/import/noop_import.dart
@@ -1,10 +1,31 @@
 // noop_import.dart — import a NOOP raw-sensor CSV export into the local store.
 //
 // The NOOP export is LONG-FORMAT raw 1 Hz: one row per decoded sample, with a
-// `stream` discriminator (hr / rr / gravity / spo2 / skintemp / resp / event) and
-// only that stream's columns filled. Header (line 4):
+// `stream` discriminator and only that stream's columns filled. Columns are read
+// by NAME from the header, never by fixed position — NOOP has already shipped one
+// schema change (see below) and name-keyed reads absorbed it without misparsing.
+//
+// SCHEMA AS SHIPPED (observed 2026-08, NOOP 9.1/9.2 — OpenStrap/edge#160):
 //   unix_s,iso_utc,stream,hr_bpm,rr_ms,grav_x,grav_y,grav_z,step_counter,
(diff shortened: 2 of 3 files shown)


abdulsaheel:
Went through every bot finding against the actual code rather than taking them at face value. **One was real and is now fixed in `533ada3`; three were false; one is acknowledged but not worth acting on.**

## ✅ Real — and worse than described (`533ada3`)

**"Idempotence Broken"** (PR Agent)

The stated premise is wrong: `hasLiveCoverageWindow` **is** an exact match, not the range check it hypothesised —

```sql
SELECT 1 FROM live_coverage WHERE start_ts = ? AND end_ts = ? LIMIT 1
```

But the failure it predicted *from* that wrong premise is real for a different reason, and it reproduced on the first try. An exact-window check only defends a **byte-identical** re-import. The realistic case walks straight past it: export again later over a longer span covering the same session.

```
import #1 (20 min):              steps=1199
import #2 (40 min, OVERLAPS #1): steps=2399
window 1785488400..1785489599 steps=1199
window 1785488400..1785490799 steps=2399
TOTAL banked = 3598   (TRUTH = 2399)   ← 50% inflated
```

The run boundary moves, no exact window matches, the whole overlap is banked twice. Exporting again next week is a completely ordinary thing to do, so this would have hit real users.

**Fix:** `stepRuns` now takes the spans already in `live_coverage` and skips any per-second delta intersecting one, breaking the run there. Clipping is **exact, not pro-rated** — we hold the counter at every second, so uncovered sub-intervals are summed from real deltas. The exact-window check is dropped, since covered-clipping strictly subsumes it. It also stops an imported span double-counting against a **live 100 Hz pedometer** window, which shares this table.

```
import #2 (40 min, OVERLAPS #1): steps=1200
window 1785488400..1785489599 steps=1199
window 1785489599..1785490799 steps=1200
TOTAL banked = 2399   ← truth
```

Real #160 export re-verified: still **2,572** on first import, **0** on re-import, one window. Four new tests (three pure, one end-to-end) pin 2,399 rather than 3,598.

## ❌ Not real — verified against code

**"Missing Absent-Input Guard"** (PR Agent) — hedged on whether `at()` can throw. It cannot:

```dart
String at(List<String> f, String name) {
  final i = idx(name);
  return (i != null && i < f.length) ? f[i] : '';
}
```

Returns `''` on any miss. A future schema dropping `step_counter` yields `int.tryParse('') == null` and the sample is skipped — the behaviour the finding asks for is already there.

**"Flush orphan-date steps before finalization"** (PR Agent suggestion) — the suggested diff adds `await engine.finalizeImport(profile);` after the flush loop. That is already the order:

```
269:  stepsBanked += await _flushStepCoverage(e.value, e.key);   ← flush
272:  await engine.finalizeImport(profile);                       ← then finalize
```

Misread of the diff; no change needed.

**"Steps Lost on Derive-Only Date"** (PR Agent) — the premise is right (a derived date is already `remove`d, so the loop only sees dates that never derived) but the conclusion doesn't follow. That loop is a deliberate backstop for a date buffered out-of-order or one where `secs` is empty at EOF. Those steps are banked to `live_coverage` and picked up whenever that day *is* derived — dropping them instead would be the actual data loss. Not "misleading": `NoopImportResult.steps` is documented as steps *banked*, not steps *derived*.

## ⚠️ Acknowledged, not acting

**TOCTOU race on the idempotency guard** (PR Agent suggestion, self-rated Low) — correct in principle. In practice the importer is strictly sequential and gated by `_busy`/`_picking` in `import_screen.dart`, so there is no concurrent path to race. Worth revisiting only if imports ever go parallel; the covered-clipping change also shrinks the window, since the check now reads spans rather than one exact key. Adding a `UNIQUE(start_ts, end_ts)` constraint would need a migration on a table the live pedometer writes to, which is a bigger change than the risk justifies here.

---

CodeRabbit was still processing when I wrote this — happy to do another pass once it lands.

Full suite after the fix: **1103 passing**, `flutter analyze` clean.

abdulsaheel:
Follow-up on the post-`533ada3` suggestion. **Checked it; it's false on both counts — but it pointed at a property worth pinning, so `b0d3526` adds a test for it.**

## ❌ "Make step coverage flush atomic via a transaction"

**The suggested API doesn't exist.** `LocalDb.inTransaction` isn't a thing in this codebase — `grep` returns nothing. The pattern here is inline `db.transaction((txn) async {...})`.

**The failure mode doesn't happen either.** The claim is that an interrupted flush leaves `live_coverage` partially written, and covered-clipping then "silently loses the unwritten steps on any subsequent re-import."

That inverts how the clipping works. Runs are **disjoint time spans**. A run that never landed is, by definition, *not covered* — so the next import banks it. Tested by writing a two-run import, deleting the second row to simulate a crash between runs, then re-importing:

```
full import:                       steps=1198
  window 1785488400..1785488999 steps=599
  window 1785489600..1785490199 steps=599
simulated partial flush: deleted run B (599 steps)
re-import after partial flush:     steps=599     ← exactly the lost run
windows=2 total=1198 (truth 1198)  ← RECOVERED
```

Not just "not lost" — recovered *exactly*, with no double-count, because the surviving run is still covered and gets clipped.

This is a direct consequence of the `533ada3` change. Keying idempotency by **time span** instead of by an exact window row is what makes the flush **recoverable rather than atomic** — which is a strictly better property than the transaction would have bought, since a transaction protects against a torn write but still leaves you with nothing banked, whereas this recovers on the next import either way.

## What I did take from it

The suggestion was wrong, but "what happens if the flush is interrupted?" is a fair question that the code only answered implicitly. `b0d3526` makes it explicit with a regression test, so nobody has to re-derive it — and so a future change that swaps time-span clipping back to an exact-window key fails loudly instead of silently reintroducing the hazard.

---

Scoreboard across both bot passes: **1 real finding** (the overlap double-count — genuinely valuable, would have hit users), **4 false**, **1 acknowledged-but-not-worth-acting-on**. The real one is fixed and the false ones are documented above rather than silently dropped.

**17 tests** in `noop_schema_drift_test.dart`, full suite **1104 passing**, `flutter analyze` clean.

PR #180, opened 2026-08-03
Stop replayed chargingOn events re-notifying (#179)

### LINKED ISSUE #179: Android App: Repeated notification "Your band is on the charger" well after the charger has been removed
### What happened?

Charged the band early today and receivign repeated notification that the charger is still on the band.

### Steps to reproduce

1. Place the charger on the band
2. Receive the initial notification
3. Notifications keep coming after charger is removed

### App version

0.9.21

### Platform

Android (APK)

### Phone model and OS version

Google Pixel 10 Pro XL

### Which band?

WHOOP 4.0

### Screenshots

_No response_

### Quick checks

- [x] The official WHOOP app is not also connected to this band
- [x] I've searched existing issues for this

abdulsaheel:
### **User description**
Fixes #179 — repeated "Your band is on the charger" long after the charger was removed (Android, 0.9.21).

## Root cause

The charging alert fires from a band **event**, and nothing on that path checked *when* the event happened.

1. `ble_engine.dart` `_absorbState` sets `state.charging` from a decoded event and calls `onState`. The decoded event carries `ts_epoch`, and it was discarded.
2. `app_state.dart` `_onEngineState` feeds every `onState` into `DeviceAlerts`.
3. `device_alerts.dart` fired on any `charging == true && _wasCharging != true`, with `_wasCharging` held **only in RAM**.
4. `notification_service.dart` re-posts id 1002 with no `onlyAlertOnce`, so each re-post re-buzzes.

The strap buffers its event log in flash, dumps it on connect, and re-sends events it has already delivered — so step 1 hands step 3 a `chargingOn` from hours ago and it is treated as a live plug-in. And because the edge state is RAM-only, **every Android process restart re-arms it**: `EdgeApplication` pre-warms a Dart engine on every process create, and `KeepAliveWorker` + the `START_STICKY` foreground service recreate the process routinely.

### Evidence

Queried a real user export (`openstrap_export_*.db`, 6,980 events):

| Check | Result |
|---|---|
| Events delivered >10 min after they occurred | **750** |
| Events delivered >1 h after | **410** |
| Worst case | **8h 26m** stale (8-hour backlog dumped in ~90 s) |
| Duplicate `(event_id, ts)` deliveries | **140 groups, up to 4×** — different frame `seq` each time, so nothing upstream de-dupes them |

The codebase already knew both facts: `db.dart`'s `insertEvent` comments *"the band re-sends events"* (hence its `ConflictAlgorithm.ignore`), and `GestureDispatcher` carries a recency window plus a debounce for exactly this — *"older than this = a drained/historical tap"*. Double-tap got that treatment; charging never did.

## The fix

**`ChargeAlertPolicy`** (new, pure, unit-tested) — two guards, because neither is sufficient alone:

- **Recency** — a `chargingOn` describing a plug-in from hours ago is history, not news. Catches first-time delivery of a backlogged event, which has no prior identity to match against.
- **Identity**, persisted — never announce a charge session at or before the last one announced. Catches exact re-sends and restart-driven re-arming, both of which are recent enough to pass the recency check.

Supporting changes:

- **`DeviceState.chargingTs`** carries the event's own strap timestamp. The flag itself still means "latest known charging state" and the UI should keep showing it — only consumers that treat the *transition* as live gate on the age.
- **`DeviceAlerts` persists its edge state**, so "once per real event" survives a process restart. The low-battery hysteresis had the identical defect (re-armed on every restart) and is fixed with it.
- **`chargingOff` cancels the Charging card.** It is a state claim, so leaving it in the tray after the puck comes off was wrong independently of this bug.
- **`onlyAlertOnce`** on the device channel — explicitly **not** load-bearing. It only suppresses sound while a card with that id is still showing, so the user dismissing it re-arms the buzz. It just stops a redundant *update* from making noise; the real de-dupe is the policy.

## Trade-off worth flagging

A genuine plug-in whose event is first delivered **more than 15 min late** no longer notifies. Real first-delivery lags in the export were 0 s, 105 s and 226 s (the app reconnecting minutes after the puck went on), so the window clears them with room to spare. A tighter window would have silenced those real alerts and fixed nothing, since re-sends are recent by definition.

A more precise alternative — debounce the charging edge until the event burst goes quiet, then act on the *resolved* final state — would preserve even a very-late first notification while still dropping a replayed on/off pair. It needs a timer and has its own failure mode (an `off` still queued for a later connection), so I left it out of a bug fix. Noted as a possible follow-up.

## Not changed, deliberately

- **Engine `state.charging` still applies old events.** Applying the replayed log in order converges to the correct current state; gating it would make the UI go blank on a legitimate plug-in the app learned about late, since there is no live charging source to fall back on.
- **HELLO's charging bit is parsed but unwired** (`control.dart:295` sets `HelloInfo.charging`; `_absorbState` reads only `batteryPct`/`wristOn` from it). A request/response bit would be a genuinely live signal, but the HELLO battery-block offsets are heuristic and unverified on hardware — not something to wire up in a bug fix. Worth a look separately.

## Testing

`test/device_alerts_test.dart` — 23 new tests: policy table, the exact #179 loop (announce → remove → 4 re-sends → 5 process restarts ⇒ still one notification), the real measured lags still announcing, a genuinely new session after a restart still announcing, unset-RTC fallback, identity expiry so a backwards strap clock can't silence alerts forever, and low-battery hysteresis surviving a restart.

Each was checked to **fail against the old behaviour** — temporarily restoring the original rule fails 7 of them, including both #179 regression tests, and removing the restore guard fails the broken-store test.

- `flutter analyze` — clean
- `flutter test --concurrency=1` — **1127 passed** (1104 + 23)
- No `kAlgoVersion` bump: notification layer only, no analytics output change.

Not verified on hardware — the policy is fully covered by unit tests, but `onlyAlertOnce` and the card-cancel are OS behaviours that need a device check: plug in → buzz; re-post while showing → silent update; unplug → card clears; plug in again → buzzes.

___

### **PR Type**
Bug fix, Tests

___

### **Description**
- Fixes repeated "band is on the charger" notifications from replayed stale BLE events (#179)

- Adds `ChargeAlertPolicy` with recency + identity guards to filter backlog replays

- Persists charging/low-battery edge state to `SharedPreferences` so Android process restarts don't re-arm alerts

- Adds 377-line test suite covering all policy paths and restart scenarios

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["BLE event\n(chargingOn + ts_epoch)"]
  B["DeviceState.chargingTs\n(strap timestamp)"]
  C["DeviceAlerts.onDeviceState\n(serialized async queue)"]
  D["ChargeAlertPolicy.evaluate\n(recency + identity guards)"]
  E["SharedPreferences\n(persisted edge state)"]
  F["NotificationService\n(idCharging / idLowBattery)"]

  A -- "ts_epoch carried" --> B
  B -- "chargingTs passed" --> C
  C -- "restore from" --> E
  C -- "evaluate verdict" --> D
  D -- "announce / stale / alreadyAnnounced / cooldown" --> C
  C -- "on announce: show/cancel" --> F
  C -- "write back" --> E
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><details><summary>5 files</summary><table>
<tr>
  <td><strong>ble_engine.dart</strong><dd><code>Carry strap event timestamp into DeviceState.chargingTs</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/180/files#diff-6ab6bfb845fca8d59abbc4cb3d3afc4d41e6ccd63608a71bb397ba7a748c7b85">+6/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>charge_alert_policy.dart</strong><dd><code>New pure policy: recency and identity guards for charging alerts</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/180/files#diff-c3f184b96c6ce67a2ec07375f38e6598a277cc21a05384d935cbcb28ee3d801f">+148/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>device_alerts.dart</strong><dd><code>Persist edge state, serialize async queue, apply ChargeAlertPolicy</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/180/files#diff-240becf3d6bb44d1bd155dcb567166f167cd4944c7d437242f09b5dc06da6bd6">+177/-17</a></td>

</tr>

<tr>
  <td><strong>notification_service.dart</strong><dd><code>Add onlyAlertOnce for device notification category on Android</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/180/files#diff-67af22f84b6b4b271b1b4d15a7b075f31e1c4417f556d99c0c1cd63950f9f5a7">+8/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>app_state.dart</strong><dd><code>Pass chargingTs through to DeviceAlerts.onDeviceState</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/180/files#diff-d56ba858e512e18dc3962da198adee46d354aca39ff7c1b4a11d5f19b9afc1ae">+5/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td><strong>models.dart</strong><dd><code>Add chargingTs field to DeviceState</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/180/files#diff-b0870ab2a85dc22705cb348d05a8e82e65cd2b495ce7852e9a5d317b77224d6d">+12/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>device_alerts_test.dart</strong><dd><code>Full test suite for charging and low-battery alert de-dupe</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/180/files#diff-5dbca7b29cb8cfa11cdaf9c43b87de39e731d7e75379068bd5e75d13bd073185">+377/-0</a>&nbsp; </td>

</tr>

</details>

___

## Summary by CodeRabbit

* **New Features**
  * Charging alerts now use event timestamps to prevent duplicate, stale, or replayed notifications.
  * Alert state persists across app restarts, including charging sessions and low-battery status.
  * Removing a device from charging cancels its active charging notification.
  * Low-battery alerts provide improved recovery and threshold handling.

* **Bug Fixes**
  * Prevented repeated notification sounds when existing device alerts are updated.
  * Improved resilience when notification or alert-state storage operations fail.

### DIFF
--- a/lib/ble/ble_engine.dart
+++ b/lib/ble/ble_engine.dart
@@ -1965,6 +1965,12 @@ class BleEngine {
     }
     if (f.containsKey('charging')) {
       state.charging = f['charging'] as bool;
+      // Carry the EVENT's own strap timestamp alongside the flag. The strap
+      // dumps its buffered event log on connect and re-sends events it has
+      // already delivered, so a chargingOn frame is not evidence that the puck
+      // went on just now — only its timestamp is. Consumers that treat the
+      // transition as live (DeviceAlerts) gate on this; see #179.
+      state.chargingTs = (f['ts_epoch'] as num?)?.toInt();
       onState(state);
     }
     if (f.containsKey('on_wrist')) {
--- a/lib/data/models.dart
+++ b/lib/data/models.dart
@@ -126,6 +126,18 @@ class DeviceState {
   String? serial;
   double? batteryPct;
   bool? charging;
+
+  /// Strap timestamp (unix sec) of the chargingOn/chargingOff EVENT that last
+  /// set [charging] — null when unknown.
+  ///
+  /// [charging] alone cannot distinguish "the puck just went on" from "the strap
+  /// replayed a chargingOn out of its flash backlog hours later" (see
+  /// charge_alert_policy.dart). It is deliberately kept beside the flag rather
+  /// than folded into it: the flag is still the LATEST KNOWN charging state and
+  /// the UI should show it, while anything that treats the transition as a live
+  /// event (notifications) has to check the age first.
+  int? chargingTs;
+
   bool? wristOn;
   int? liveHr; // latest live HR from the foreground stream
   int? liveHrAt; // epoch ms
--- a/lib/notify/charge_alert_policy.dart
+++ b/lib/notify/charge_alert_policy.dart
@@ -1,0 +1,148 @@
+// charge_alert_policy.dart — decides whether a "charging started" band event is
+// NEWS (worth buzzing the user) or NOISE (a backlog replay of something that
+// already happened).
+//
+// WHY THIS EXISTS (issue #179 — "repeated 'band is on the charger' well after
+// the charger was removed"):
+//
+// The strap buffers its event log in flash and dumps it on connect, and it
+// re-sends events that were already delivered. Both are measurable in a real
+// user export: 410 of 6,980 events arrived more than an hour after they
+// occurred (worst case 8h26m, an 8-hour backlog dumped in ~90 s), and 140
+// distinct (event_id, ts) pairs were delivered more than once — up to 4× —
+// with a different frame `seq` each time, so nothing upstream de-dupes them.
+// `db.dart`'s insertEvent already notes "the band re-sends events", and
+// GestureDispatcher already carries a recency window for exactly this reason
+// ("older than this = a drained/historical tap"). The charging alert never got
+// the same treatment: it fired on any false→true edge, with the edge state held
+// only in RAM — so every Android process restart (EdgeApplication pre-warm,
+// KeepAliveWorker, FGS START_STICKY) re-armed it and the next replayed
+// chargingOn buzzed again.
+//
+// Two independent guards, because neither is sufficient alone:
+//   • RECENCY — an event describing a plug-in from hours ago is not news.
+//     Handles first-time delivery of a stale backlogged event.
+//   • IDENTITY — never announce a charge session at or before the last one
+//     already announced, persisted across process restarts. Handles exact
+//     re-sends and restart-driven re-arming, both of which are recent enough
+//     to pass the recency check.
+//
+// Pure and synchronous: the caller owns the clock, the persisted values and the
+// notification plugin. See device_alerts.dart for the wiring.
+
+/// Why a charging event did (or did not) raise an alert. Returned instead of a
+/// bare bool so the caller can log the reason — the failure this fixes was
+/// invisible in logs precisely because nothing recorded WHY a notification
+/// fired.
+enum ChargeAlertVerdict {
+  /// Genuine new charge session — show the notification.
+  announce,
+
+  /// We already believe we are charging (in-process edge state) — the strap
+  /// re-sent chargingOn, or another field on the same DeviceState changed and
+  /// re-delivered the unchanged charging flag.
+  alreadyCharging,
+
+  /// The event describes a plug-in that happened long enough ago that it is
+  /// history, not news — a flash-backlog replay.
+  stale,
+
+  /// This exact charge session was already announced (same or older strap
+  /// timestamp than the last announcement) — a re-send, or a fresh process
+  /// re-processing an event the previous process already alerted on.
+  alreadyAnnounced,
+
+  /// The event carries no usable timestamp AND we announced recently enough
+  /// that another alert would be a duplicate.
+  cooldown,
+}
+
+class ChargeAlertPolicy {
+  const ChargeAlertPolicy._();
+
+  /// How recent a chargingOn event must be to count as news.
+  ///
+  /// Deliberately generous. The point is to reject the HOURS-stale backlog
+  /// dump, not to demand instant delivery: in the same real export, genuine
+  /// first-delivery charging events lagged 0 s, 105 s and 226 s behind the
+  /// strap timestamp (the app reconnecting a few minutes after the puck went
+  /// on). A tight window would silence those real alerts and fix nothing —
+  /// exact replays are caught by the identity check below, not by this one.
+  static const int liveWindowSec = 900; // 15 min
+
+  /// Above this age a timestamp is not evidence of anything — the strap ships
+  /// with its RTC unset and can wander before SET_CLOCK lands. We refuse to
+  /// judge staleness from it (mirrors GestureDispatcher's `_plausibleAgeCapSec`)
+  /// and fall back to the untimed path.
+  static const int implausibleAgeSec = 86400;
+
+  /// Below this a strap timestamp is clearly a wrong/unset RTC rather than a
+  /// real unix time — the same plausible-unix floor the sync clock policies use.
+  static const int plausibleUnixFloor = 1700000000;
+
+  /// Tolerance for a strap clock that reads slightly AHEAD of the phone. Inside
+  /// this the event is still "now"; beyond it the timestamp is not usable.
+  static const int futureSlackSec = 300;
+
+  /// Minimum gap between alerts when the event carries no usable timestamp, so
+  /// an unset-RTC strap can't buzz on every reconnect.
+  static const int untimedCooldownSec = 3600; // 1 h
+
+  /// The identity check is not applied to an announcement older than this.
+  ///
+  /// Without an expiry, a strap clock that jumps BACKWARDS (into a still-
+  /// plausible range, so the floor above doesn't catch it) would leave every
+  /// future session comparing `<=` against a high-water timestamp that never
+  /// gets beaten — charging alerts would go permanently silent, a worse and far
+  /// less visible bug than the one being fixed. Expiring the comparison costs
+  /// nothing: a replay old enough to reach this point is already rejected by
+  /// the recency window.
+  static const int identityExpirySec = 43200; // 12 h
+
+  /// Whether [tsEpoch] can be trusted as a real strap wall-clock reading.
+  static bool timestampUsable(int? tsEpoch, {required int wallNowSec}) {
+    if (tsEpoch == null || tsEpoch < plausibleUnixFloor) return false;
+    final age = wallNowSec - tsEpoch;
+    if (age < -futureSlackSec) return false; // strap clock ahead of the phone
+    return age < implausibleAgeSec;
+  }
+
+  /// Decide whether a chargingOn edge should raise a notification.
+  ///
+  /// [eventTsEpoch] is the strap timestamp of the event that set `charging`
+  /// (null when unknown). [wasCharging] is the in-process edge state (null =
+  /// nothing seen yet this process). [lastAnnouncedEventTs] and
+  /// [lastAnnouncedWallSec] are the persisted results of the previous
+  /// announcement — they are what survives an Android process restart.
+  static ChargeAlertVerdict evaluate({
+    required int? eventTsEpoch,
+    required int wallNowSec,
+    required bool? wasCharging,
+    required int? lastAnnouncedEventTs,
+    required int? lastAnnouncedWallSec,
+  }) {
+    if (wasCharging == true) return ChargeAlertVerdict.alreadyCharging;
+
+    if (!timestampUsable(eventTsEpoch, wallNowSec: wallNowSec)) {
+      // No usable strap clock → we cannot tell news from replay. Fall back to a
+      // blunt wall-clock cooldown rather than guessing.
+      if (lastAnnouncedWallSec != null &&
+          wallNowSec - lastAnnouncedWallSec < untimedCooldownSec) {
+        return ChargeAlertVerdict.cooldown;
+      }
+      return ChargeAlertVerdict.announce;
+    }
+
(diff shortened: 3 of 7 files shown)


abdulsaheel:
Reviewed both bots against the code. **4 of 7 findings were real and are fixed in 5710526**; 3 declined with reasons below.

### Fixed

**1. Duplicate plausible-unix floor** (CodeRabbit) — real. `kMinPlausibleUnix = 1700000000` already exists in `sync_policy.dart:55`, and `ble_state.dart:13` already imports from there. Now imported rather than re-declared.

**2. The 24 h usability cap was itself a bug** (surfaced by CodeRabbit asking for a test at `age > implausibleAgeSec`). Writing that test showed the behaviour it would have pinned is wrong: the cap was modelled on `GestureDispatcher._plausibleAgeCapSec`, which is right for a tap (bounded backlog) and wrong here — the strap banks *days*, so a two-day-old `chargingOn` is a real old event, not a broken clock. The cap routed it to the untimed path where it **announced**, reintroducing #179 past 24 h. Cap removed; an unset RTC reads near zero and is still caught by the floor. The test pins the fix instead of the bug.

**3. A replayed `chargingOff` must not clear a live Charging card** (CodeRabbit) — real. Writing the test also exposed a second defect in my own first attempt: I keyed the cancel off the charging *transition*, but a stale `chargingOff` still has to update `_wasCharging` (it is the latest known state), which **consumed the transition** — so the real removal that followed found nothing to act on and stranded the card in the tray. The cancel now keys off a separate "already cancelled for this off-state" latch. Both paths covered.

**4. Latch ordering** (PR Agent, citing AGENTS.md §4.3) — real, and the citation checks out: §4.3 "Sticky boolean latches never reset on the failure path" is a documented recurring class in this repo. `_apply` is now decide-then-act, every latch committed before the first `await`. A throwing sink previously left `_lowArmed` armed and retried on every update; a throwing store could re-announce the same session — and the queue's `catchError` made both silent. Restore also assigns the identity pair atomically.

### Declined

**Route device alerts through `NotificationCenter.emit`** (CodeRabbit). Checked what it would actually add for this category: `notification_prefs.dart:120` reads `NotifCategory.device => true, // device alerts aren't user-gated here`, so no category toggle is being bypassed — the only added gate is **quiet hours**. That would silence a confirmation of an action the user just performed, during exactly the window in which people put a band on the charger (before bed). Its fire-once guard is also set-membership on a `dedupeKey` via `FiredKeyStore`, which cannot express the `<=` identity comparison this needs. I don't think §4.6's "bypassing the prefs gate" applies when the prefs gate is a literal `true` for this category — but happy to revisit if a device toggle ever lands.

**Persist-before-announce durability gap** and **untimed cooldown with no prior wall record** (PR Agent). Both require a SharedPreferences write to fail. In that window the recency check still rejects anything stale, and after this change the in-memory latches cover the same session regardless of whether persistence succeeded. Narrow enough that I'd rather not add code for it.

One correction to a PR Agent suggestion I *did* partly take: it argued "the real risk is the inverse — `_lastAnnouncedWallSec` set but `_lastAnnouncedEventTs` null". That specific direction is unreachable, because `_kLastChargeTs` is read first. I made the assignment atomic anyway so the invariant is local rather than dependent on statement order, and noted why in the comment.

`flutter analyze` clean; **1133 tests** (1127 → 1133). Each new test was confirmed to fail against the pre-fix code — restoring the 24 h cap fails 2, and moving the low-battery disarm back after the `await` fails the latch test.

abdulsaheel:
Second round reviewed. **1 of 4 real, fixed in a32e92f.**

### Fixed

**Untimed announcement left a stale identity high-water** (PR Agent) — real, and the failure direction is the bad one. A session announced with no usable strap clock has no identity of its own, but the previous high-water stayed in place and went on to judge a session it knew nothing about: the next usable timestamp falling at or below it got suppressed, costing the user a real "on the charger" confirmation. That's worse than the duplicate this PR exists to kill. Now cleared in memory and persisted as `0` (every real strap timestamp is far above it), so the untimed path is guarded only by the wall-clock cooldown — which is what `ChargeAlertPolicy`'s untimed branch already assumes. Test seeds an old high-water, announces untimed, then announces a genuinely-new-but-lower-timestamped session; verified to fail without the fix.

### Declined

**"Low-armed not persisted on restore"** and **"untimed cooldown not restart-safe"** — both scenarios require SharedPreferences writes to fail. Not a regression: before this PR `_lowArmed` was RAM-only and re-armed on *every* process start, so a dead store degrades to exactly the old behaviour while a working one is strictly better. There is no way to remember a latch without somewhere to remember it. I did note the asymmetry in a comment on `_kLowArmed`: charging keeps the recency window as a second line of defence, but a battery percentage carries no timestamp, so low-battery has only the key. `_ThrowingStore` is a fixture I wrote for robustness, not an observed production condition.

**"Remove `_cancelledForOff` — it blocks real card cancellation"** — the stated mechanism doesn't hold. The claim is that "a fresh process receiving a replayed **stale** chargingOff will call `_notes.cancel` … and set `_cancelledForOff = true`, which then blocks the cancel for the real removal". A stale chargingOff can't reach either statement — `!ChargeAlertPolicy.isStale(...)` gates the whole expression, which is the fix from round 1. Only a non-stale off sets the latch, and that *is* the real removal.

On the underlying question — should the latch be persisted? No: dropping it means calling `cancel()` on every update where `charging == false`, and `state.charging` stays false for the rest of the process once a chargingOff is seen, while `_onEngineState` fires at ~1 Hz on live HR. That's a platform round-trip every second, forever. Its failure mode after a restart is one *redundant* cancel — a no-op — never a missed one, i.e. the opposite of the claim. The existing "charging off clears the card even on a fresh process" test covers that path.

`flutter analyze` clean; **1134 tests**.

abdulsaheel:
Third round. **1 of 2 real, fixed in 04dd2bf.**

**Presentation could take persistence down with it** (PR Agent) — real. A throwing `_notes.show` aborted `_apply` before the `_store.writeInt` calls below it, so a session could be shown but never recorded, and would re-announce after a restart.

Worth noting the suggested reorder wouldn't have fixed it — persist-then-show just swaps which failure is fatal, letting a dead store swallow the alert entirely. Since neither step reads the other's result, each is now attempted independently and absorbs its own failure.

**"Low-battery re-arm on stale chargingOn"** — the analysis walks through it and reaches "No actual bug here", which matches what I get: a stale `chargingOn` re-opens `_cancelledForOff`, and the stale `chargingOff` that follows is still stopped by `isStale`. Agreed on the fragility point though — nothing asserted that interaction, so it's now pinned in "a stale on/off pair does not swallow the next genuine plug-in".

`flutter analyze` clean; **1135 tests**. Both new assertions verified to fail without the change.

---

Heads-up on the check status: CodeRabbit's green tick on the last two pushes reads **"Review rate limited"** — it never analysed those commits, so its pass carries no signal here. Every finding actioned in rounds 2 and 3 came from PR Agent. Worth a re-run before merge if you want a second opinion on the latest diff.

abdulsaheel:
Fourth round — **no action; this re-raises a finding the same reviewer resolved last round**, and the new mechanism doesn't hold.

The claim is that a stale `chargingOff` "is NOT stale by the isStale check because its ts is also old" and will therefore cancel the card. That's self-contradictory — an old timestamp is exactly what `isStale` returns true for:

```dart
static bool isStale(int? tsEpoch, {required int wallNowSec}) =>
    timestampUsable(tsEpoch, wallNowSec: wallNowSec) &&
    wallNowSec - tsEpoch! > liveWindowSec;
```

The behaviour is asserted, not argued — `a stale on/off pair does not swallow the next genuine plug-in` feeds `on(9h old)` then `off(8h old)` and checks `sink.cancelled` does **not** contain `idCharging`. Added in 04dd2bf, passing.

The second half asks for coverage of the low-battery re-arm side, which already exists as `a replayed plug-in does not re-arm the low alert` — cited in the previous round's own analysis. Both pass:

```
DeviceAlerts charging a stale on/off pair does not swallow the next genuine plug-in  ✓
DeviceAlerts low battery a replayed plug-in does not re-arm the low alert            ✓
```

Calling the automated review done here: rounds 2–4 have converged on re-litigating settled ground, and CodeRabbit has been rate-limited (not actually analysing) for the last three pushes. Six real findings were fixed across the three rounds; the rest are documented above with reasons. Ready for human review.

PR #182, opened 2026-08-03
Steps from a real pedometer only; movement minutes on measured evidence

abdulsaheel:
## Why

A user reported **2,645 steps on a day they took under 400**. Diagnosed on their real DB: **81.5% of every step figure this app has ever shown was manufactured** — 41,753 of 51,225 steps across 26 days — by a 1 Hz estimator that physically cannot see gait.

| day | before | after | source |
|---|---|---|---|
| 08-03 | 2,645 | **—** | none |
| 08-02 | 7,601 | 11 | band |
| 08-01 | 6,900 | **—** | none |
| 07-29 | 6,657 | 907 | band |
| **total** | **51,225** | **9,472** | |

7 of 26 days now correctly show nothing. They fill in with real counts once phone steps are enabled.

## Steps are real-measured only (`kAlgoVersion` 54 → 56)

`scalars.steps` is **absent, not 0**, unless a gait-capable source measured the day. The old hybrid persisted a hard `0.0` whenever the estimator abstained, which poisoned the `dyn_p90` median and dragged every average and "most steps" record down.

## Phone pedometer

The band is on the wrist and its 24/7 stream is 1 Hz, where walking is unresolvable. The phone rides in a pocket, sees trunk motion, and already counts steps into the on-device health store.

- **No new dependency** — `health` was already in pubspec, with HealthKit entitlements, `NSHealthShareUsageDescription` and `READ_STEPS` already declared.
- Uses `getTotalStepsInInterval`, which on iOS is an `HKStatisticsQuery` cumulative sum — HealthKit de-duplicates iPhone/Watch overlap itself, which a raw sample read would not.
- Everything stays on-device; nothing is uploaded.

**`live_coverage` gains a `source` column** (db v26 → v27, via the existing guarded `_addColumnIfMissing`). Phone and band counts are **never summed** — they're the same walk seen from pocket and wrist, so adding them roughly doubles a day. Phone wins outright when present; band is the fallback. Sync is delete-then-insert scoped to `source='phone'`, idempotent by construction. Disabling the toggle drops the phone rows, since a stale row would otherwise keep overriding the band from a source no longer being read.

**Steps are no longer written to Apple Health / Health Connect.** The old value was fabricated, and we now *read* that store — writing our copy back would double-count into it and feed our own number back to us. `STEPS` stays in the export type list so the per-day delete pass **actively purges the samples earlier versions wrote**.

## Movement minutes — every change proven on real substrate first

| Change | Evidence |
|---|---|
| **HR gate deleted** | Changed active minutes by **0** on every day tested. At RHR ~62 it sits at **~6% of heart-rate reserve** (below every ACSM band); 73–100% of covered minutes already cleared it. Failed in the wrong direction too — PPG HR is least reliable during the motion being gated. |
| **×3 ceiling deleted** | Rejected **0 minutes**, 0.42–0.55 g headroom. Cannot fire on artifacts (a 3 s knock averages ~0.23 g, below the *floor*). |
| **Floor frozen** after 14-day enrollment, persisted in `baselines` | See table below |
| **Coverage exclusion dropped** | It existed only to stop step double-counting; there is no longer a step total to double-count into. |

| activity | frozen floor | recomputed floor |
|---|---|---|
| 1.0× | 23 | 37 |
| 2.0× | 128 | 37 |
| 3.0× | 254 | **37** |

A recomputed floor reports the same number whether the user tripled their activity or did nothing. Re-freezes only on device/wrist change, a 30-day wear gap, or 365 days.

## Refuted — deliberately not built

- **Sleep-anchored floor**: CV **138.6%** across days vs 9.3%; on one night it landed above the entire day's range (would report zero). Tested twice, including with real detected sleep windows.
- **Accel autocalibration**: +5% gain moves the gate decision by **0.0000**.
- **Gravity orientation**: solved the ambulation problem this deletes.

## Root cause, finally

The R24 1 Hz accel field is a **fused gravity vector, not acceleration** — p50 **1.027 g** across 269,486 samples; **1.033 g ± 0.006** during the most vigorous minute of a day, 0 of 420 samples above 1.2 g. So ENMO over this substrate reduces to `~(1.03 − gRef)`: a pure calibration artifact with zero signal. That is why the old estimator gave 42,155 steps at gRef 0.97 and 0 at 1.02.

## Caveat

All measurements rest on **4 days** of per-minute substrate (the export's retention limit). The direction of each result is unambiguous; the CV figures are thin.

## Test plan

- `flutter analyze` clean
- `flutter test --concurrency=1` → **1118 passing**
- **CI reproduced locally against the pinned analytics** (`pubspec_overrides.yaml` moved aside, `resolved-ref` verified) before committing — analyze + full suite both green
- `pubspec.lock` diff verified to be *only* the pin change, no `path:` pollution
- New: 8 tests on phone/band source preference and double-count paths, 6 on frozen-floor persistence

___

### **PR Type**
Bug fix, Enhancement, Tests

___

### **Description**
- Steps are now real-measured only: the fabricated 1 Hz wrist estimate is deleted; `scalars.steps` is absent (not zero) unless a gait-capable source (band 100 Hz or phone pedometer) measured the day

- New `PhonePedometer` reads hourly step counts from HealthKit/Health Connect into `live_coverage` with `source='phone'`; phone wins over band when present, the two are never summed

- `kAlgoVersion` bumped 54 → 56; schema bumped 25 → 27 (`live_coverage.source` column added, frozen movement floor stored in `baseline` table)

- Movement minutes rebuilt as a non-locomotion activity index with a frozen personal floor; fabricated steps purged from Apple Health/Health Connect export

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Phone HealthKit /\nHealth Connect"]
  B["PhonePedometer\n(new)"]
  C["live_coverage\n(source='phone')"]
  D["Band 100 Hz\nAN-2554"]
  E["live_coverage\n(source='band')"]
  F["liveStepsForDay\n(phone wins; never summed)"]
  G["_stepsAndEnergy\n(real counts only)"]
  H["scalars.steps\n(absent if no real source)"]
  I["bundle movement\n(active_min, frozen floor)"]
  J["HealthExporter\n(STEPS deleted, old samples purged)"]

  A -- "getTotalStepsInInterval" --> B
  B -- "replacePhoneCoverageForDay" --> C
  D -- "addLiveCoverage" --> E
  C --> F
  E --> F
  F --> G
  G --> H
  G --> I
  H --> J
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>derivation_engine.dart</strong><dd><code>Delete 1 Hz step estimate; bump kAlgoVersion 54→56; freeze movement </code><br><code>floor</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-0e401d8cdb1091d9a9591e9d92c4c5e979293f8480da203b22011a7cf9072ea6">+244/-104</a></td>

</tr>

<tr>
  <td><strong>db.dart</strong><dd><code>Schema v26→v27: add live_coverage.source, phone CRUD, frozen floor </code><br><code>storage</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-b96f416000a07a912224053a34f440c27993e014d324d6ae8c9dfe88bb0617e6">+136/-8</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>health_export.dart</strong><dd><code>Stop writing fabricated steps; purge old STEPS samples from health </code><br><code>store</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-7688da159b76fc36f532ed153ea690343655e60ab0436820d1bcbc54113e15f5">+23/-15</a>&nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Enhancement</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>phone_pedometer.dart</strong><dd><code>New PhonePedometer reads hourly steps from on-device health store</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-4ec05306b14837da4ca6e973be596991359aca8fe3d99c40a2875f0a13271156">+138/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>app_state.dart</strong><dd><code>Wire PhonePedometer: enable/disable toggle, sync on launch and after </code><br><code>export</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-d56ba858e512e18dc3962da198adee46d354aca39ff7c1b4a11d5f19b9afc1ae">+62/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>profile_screen.dart</strong><dd><code>Add phone step count toggle to profile health section</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-84850f115ee94849e171ca073ceabf7528d37007bea1de48e67fe0649d96e849">+48/-0</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td><strong>derive_day_window_test.dart</strong><dd><code>Update test: no-gait-source day must have null steps, not a fabricated </code><br><code>value</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-1d4315dbad09a83156a6b6a6820d34516e56b234039fe87478fafca36c5515fd">+19/-5</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>movement_floor_frozen_test.dart</strong><dd><code>New tests: frozen movement floor round-trip, thaw policy, degenerate </code><br><code>rejection</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-9d2882f9482e60d48381f4fac52bcd6084d7b870403a3fee921d36c9d86c23e6">+88/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>phone_step_source_test.dart</strong><dd><code>New tests: phone/band source priority, idempotency, clear-and-fallback</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-afb2246e8b921a2f84f748c8f76399ef9e0d137e11338da3e0b944af5f96a2a4">+129/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>step_personal_floor_test.dart</strong><dd><code>Update calls from dailyStepEstimate to dailyActiveMinutes; fix </code><br><code>assertions</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-5201e9b418ee711aa44b92d455f50d95e90297fd7181447ce5c4700f592fe58c">+5/-5</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Dependencies</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>pubspec.yaml</strong><dd><code>Temporary PR pin to analytics#35 HEAD (00a2efe); must re-pin before </code><br><code>merge</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/182/files#diff-8b7e9df87668ffa6a04b32e1769a33434999e54ae081c52e5d943c541d4c0d25">+16/-1</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

## Summary by CodeRabbit

- **New Features**
  - Added optional phone step tracking with permission controls, automatic synchronization, and recent-history updates.
  - Added phone-step settings and sync status to the health screen.
  - Phone and wearable step data now use clear source prioritization.

- **Improvements**
  - Steps are reported only from measured gait data; movement minutes remain independent.
  - Personal movement thresholds persist and update under qualifying conditions.
  - Health exports no longer create new step samples.
  - Removed step calibration; unmeasured days are clearly labeled.

- **Bug Fixes**
  - Improved handling of partial, empty, invalid, and failed phone synchronizations.


abdulsaheel:
Thanks — this was a strong review. Every finding I checked was real; fixed in `87a8a5d`.

**The most serious was the one outside the diff**: `_stepsAndEnergy`'s early returns gating the step assignment. You're right that steps now depend on nothing from the band substrate, so a day with real phone-pedometer steps but little band data reported **no steps at all** — discarding a real measurement because an unrelated signal was missing. Extracted to `_writeSteps` and hoisted above both guards.

Also fixed: the re-freeze path destroying a usable stored floor; the dead wear-gap trigger (now derived from missing `dyn_p90` days and passed — `deviceChanged`/`wristChanged` stay deliberately *unpassed* rather than fabricated as `false`, with the reason inline); the Android `hasPermissions` null→false gate, which did contradict `health_export.dart`'s own documented finding in this same codebase; the DST-unsafe hour/day arithmetic; and `healthSyncNow` writing phone rows back after the user disabled them.

**On the missing v27 migration test** — you were right that the path every real install takes was untested. Added it, and confirmed it isn't decorative by **mutation**: flipping the migration default from `'band'` to `'phone'` makes it fail 137 → 0.

Two deliberately not changed, with reasons in the commit: the STEPS purge running every pass rather than behind a one-shot cursor (real but trivial cost, and the purge is worth more than the round trip), and the temporary analytics pin — that one is intentional and flagged in both the PR body and a comment directly above the ref; it gets re-pinned to the analytics main SHA before merge.

1120 tests pass; CI reproduced locally against the pinned analytics.

abdulsaheel:
@coderabbitai review

abdulsaheel:
Checked all of these against the code before touching anything — 4 real, 6 refuted. Fixed in `f97e278`.

**The one with teeth was the all-zero phone read.** `replacePhoneCoverageForDay` is delete-then-insert, and `phone_pedometer.dart` already documents the exact failure that produces an all-zero read: on iOS `requestAuthorization` reports success even when the user denied READ, so queries come back **empty rather than null**, permanently. One sync after that and a real multi-thousand-step day was gone — and not even the band fallback showed, because phone rows win outright. An all-zero read over a day that already holds phone steps now keeps the banked day and reports it unread. Mutation-verified: the new test fails 137 → 0 with the guard stubbed out.

**Absent steps no longer label themselves.** `tier: 'ESTIMATE'` on a `value: null` block was worse than cosmetic — `Metric.parse` maps that tier to `beta: true`, so a day nothing measured rendered the estimate badge on a card with no number on it. Now `tier: null` + empty `inputs_used`. I did **not** take the suggested `'ABSENT'`: `Tier.all` in analytics is a closed set of four published grades, and the edge must not widen it from here. kAlgoVersion 56 → 57, because the persisted bundle changes even though no value does.

Also fixed: `requestPhoneSteps`/`disablePhoneSteps` set the in-memory flag *before* awaiting the pref write (a failed write left the toggle on for the run and off next launch, banking rows the restored state says were never enabled — which nothing then clears); and the `containsKey('steps')` assertion, which was **vacuous** — `got` is built by the test's own helper seeding every key unconditionally, so it could not fail whatever the derivation did. Replaced with real assertions on the persisted bundle.

**Refuted, with reasons:**

- **`_AsyncLock` "swallows the error"** — `whenComplete` *does* release the lock on the error path, and the error still propagates to the caller. The suggested `completeError` would **introduce** a bug: `_tail` would carry an error, so the next waiter's `previous.then(...)` would skip its action entirely and inherit an unrelated failure.
- **`h < 25` → `h < 26`** — `DateTime(y, m, d, 24)` normalises to `nextMidnight` in every timezone, so h=24 and h=25 both break and 26 changes nothing. The fall-back hour is already covered: bucket `[DateTime(y,m,d,1), DateTime(y,m,d,2))` spans **both** occurrences of 1am, and the walk runs to `nextMidnight`.
- **SnackBar → `NotificationCenter.emit`** — the one-emitter rule (§3.8/§4.6) is about `NotificationService` system alerts. This is an in-app SnackBar in direct response to a tap, the pattern used in 13 other UI files.
- **Floor commit on the wrong day** — `mayCommitFloorOn` is `dayId.compareTo(frozenOn) >= 0`, not `<`, and `stored` is re-read *inside* the lock. A backfill day older than a freeze a newer day just wrote is rejected.
- **Concurrent `syncPhoneSteps`** — each day is ONE transaction, so two overlapping syncs cannot interleave a delete past an insert. A re-entrancy latch would only silently drop the 7-day backfill.
- **The analytics pin** — intentional and flagged in the PR body and above the ref itself; re-pinned to the analytics main SHA once analytics#35 merges.

`flutter analyze` clean, `flutter test --concurrency=1` → **1141 passing**, `pubspec.lock` verified free of `path:` pollution.

Issue #208, opened 2026-08-06
Permenant Disconnect after not wearing/out of range


swimmerjaxon:
### What happened?

A successfully paired strap will disconnect from the app if you are not wearing it or get out of bluetooth range. The strap will refuse to reconnect to the application. The app will show it is attempting to reconnect, then simply default to 'disconnected'. If I forget/re-pair the whoop then the strap reconnects normally, or if you place the strap into blue light pairing mode the app (without forgetting) will see it and connect, syncs data and things keep moving on. 

### Steps to reproduce

Connect the strap and wear it. 

Take it off for >10min or more or walk far enough away from your phone to disconnect

put it back on

Strap/App will not reconnect unless forgotten/repaired or the strap is put into pair (blue light mode) to get reconnected to the app

### App version

0.9.21+52

### Platform

Android (APK)

### Phone model and OS version

Pixel 10 Pro

### Which band?

WHOOP 4.0

### Quick checks

- [x] The official WHOOP app is not also connected to this band
- [x] I've searched existing issues for this

cb0806151:
I'm also getting the issue above

Phone: Moto G power 2022
Android version: 12 S3RQS32.20-42-10-9-12
Edge version: v0.9.21
Band: whoop 4.0

I left my phone outside my room last night and when I try and sync it just syncs indefinitely and never reconnects.

abdulsaheel:
Fixed in [v0.9.24](https://github.com/OpenStrap/edge/releases/tag/v0.9.24).

Reconnect was edge-triggered inside a single try/catch, so one throw ended reconnection for the lifetime of the process — which matches the symptom exactly: it shows "reconnecting", gives up, and only a re-pair or blue-light mode brings it back. It is now a level-triggered supervisor that re-arms rather than depending on the edge.

Please reopen if you encounter it again.

Issue #206, opened 2026-08-06
Strain is 0.0


ramishroshan:
### What happened?

Steps that caused the issue:
1. I started a workout in the Whoop app.
2. I put the app in the background and used the Hevy app to track my workout.
3. When the workout was done, I opened the Whoop app again to stop it.
4. Because the app was in the background for a long time, it reloaded.
5. After it reloaded and I stopped the workout, the Strain was recorded as 0.0, even though heart rate data was successfully tracked.

### Steps to reproduce

<img width="645" height="1398" alt="Image" src="https://github.com/user-attachments/assets/fe238a9f-745c-4486-9de3-8ad4116b90ba" />

### App version

0.9.21

### Platform

iOS (TestFlight)

### Phone model and OS version

14 pro max iOS 27 db4

### Which band?

WHOOP 4.0

### Screenshots

<img width="645" height="1398" alt="Image" src="https://github.com/user-attachments/assets/dd1cd529-168a-4b8a-aff8-065deefa6195" />

### Quick checks

- [x] The official WHOOP app is not also connected to this band
- [x] I've searched existing issues for this

abdulsaheel:
Fixed in [v0.9.24](https://github.com/OpenStrap/edge/releases/tag/v0.9.24).

Strain was accumulated in memory by the foreground app, so whatever the app slept through was simply not in it — backgrounded long enough (or killed by iOS), a workout ended scored from only the minutes the app was awake for. The band had the whole session at 1 Hz regardless, so sessions are now rescored from that on read and after each sync. Workouts already recorded with 0.0 correct themselves once they sync.

Please reopen if you encounter it again.

Issue #200, opened 2026-08-05
Excessive background battery drain (Andriod)


juliuswms:
### What happened?

On Android, the app consumes a lot of battery in the background (~22.1% battery with 15 minutes of screen-on time). This makes daily-use and stress testing impractical.

### Steps to reproduce

1. Install app
2. Keep it running in the background

### App version

0.9.21+52

### Platform

Android (APK)

### Phone model and OS version

Samsung Galaxy S22, Android 16

### Which band?

WHOOP 4.0

### Screenshots

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/335463d1-e91a-42d7-a884-fa554c50a2f7" />
<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/470f1534-343d-4832-b0a9-fe97e38abc99" />

From looking at the code and with the help of DeepSeek I was able to understand that after the App starts, the BLE Connection is opened indefinitely. Pulling the heartbeat very frequently (10s) and pulling the banked data from the WHOOP every 15 min. My guess would be that this constant BLE connection is the reason the battery drain is this high. Maybe with a "Power save" option it could only connected every 15 min, pull the banked data and close the BLE Connection again (I guess this would come with the downside of not being able to use the Double-Tap on the WHOOP to eg. skip a song. Seeing that that would require a constant connection).
I also understand that power optimization at the start of a project like this one is not as much of a focus as new features but for me it makes daily driving this App kinda hard.

### Quick checks

- [x] The official WHOOP app is not also connected to this band
- [x] I've searched existing issues for this

gerageragarza-prog:
Algoritmo de reconexión (backoff exponencial acotado) — el que preguntaste

ReconnectBackoff.kt (implementación completa, 35 líneas, fácil de portar a C):
https://github.com/ryanbr/noop/blob/main/android/app/src/main/java/com/noop/ble/ReconnectBackoff.kt
ReconnectBackoffTest.kt (casos de prueba, útil para validar tu port a ESP32):
https://github.com/ryanbr/noop/blob/main/android/app/src/test/java/com/noop/ble/ReconnectBackoffTest.kt

Cliente BLE principal (Android) — cadencia de offload, power saving, uso del backoff

WhoopBleClient.kt (7790 líneas — usa view_range en tu editor si lo clonas, es grande):
https://github.com/ryanbr/noop/blob/main/android/app/src/main/java/com/noop/ble/WhoopBleClient.kt
[...]

Issue #199, opened 2026-08-05
NOOP and WHOOP data import both fail with FormatException (Unexpected extension byte)


gerageragarza-prog:
### What happened?

Tried importing historical data from both NOOP and WHOOP using the Import data screen. Both imports fail immediately with a FormatException, just at different byte offsets:

- Import from NOOP (raw 1Hz CSV): FormatException: Unexpected extension byte (at offset 18)
- Import from WHOOP (export CSV, BETA): FormatException: Unexpected extension byte (at offset 10)

I expected both files to import and be parsed/re-analyzed successfully, appearing alongside my existing band data as the screen describes ("Imported days sit alongside your band data").

Since both failures are the same exception type, just at different offsets, this looks like a shared decoding bug in the import pipeline (possibly a BOM or non-UTF-8 byte in the CSV being hit before parsing) rather than two separate issues.

### Steps to reproduce

1. Go to Settings > Import data
2. Tap "Import from NOOP" and select a raw 1Hz CSV export from NOOP → fails with FormatException: Unexpected extension byte (at offset 18)
3. Tap "Import from WHOOP" and select a WHOOP export CSV → fails with FormatException: Unexpected extension byte (at offset 10)
4. Both fail on the first attempt, consistently

### App version

iOS 26.5.2

### Platform

iOS (TestFlight)

### Phone model and OS version

iPhone 16 Pro – iOS 26.5.2

### Which band?

WHOOP 5.0 / MG (experimental — expect breakage)

### Screenshots

<img width="721" height="1568" alt="Image" src="https://github.com/user-attachments/assets/a86a5ea6-e36a-4648-bf72-f410c5904f1a" />
<img width="721" height="1568" alt="Image" src="https://github.com/user-attachments/assets/3a296be1-f4aa-4c2f-8365-8387de330cec" />

### Quick checks

- [x] The official WHOOP app is not also connected to this band
- [x] I've searched existing issues for this

gerageragarza-prog:
Parser principal (iOS/Mac)

WhoopExportImporter.swift (parseo de CSV, manejo de localización, zip/carpeta):
https://github.com/ryanbr/noop/blob/main/Packages/StrandImport/Sources/StrandImport/WhoopExportImporter.swift

Store writer (iOS/Mac)

WhoopImporter.swift (mapeo a tablas internas, wake-day keying, escalas):
https://github.com/ryanbr/noop/blob/main/Strand/Data/WhoopImporter.swift

[...]

PR #209, opened 2026-08-08
fix: apple health strength workouts, workout strain, steps, imports, android battery + reconnect

abdulsaheel:
Fixes a batch of open issues.

**Strength workouts never showed up in Apple Health (#184).** I was sending `STRENGTH_TRAINING` on both platforms, but HealthKit doesn't have that value — it rejected every one and the error went to a `debugPrint`. Swims had the mirror-image bug on Android.

**Strain 0.0 after a long workout (#206).** Strain was accumulated in memory by the foreground app, so whatever the app slept through simply wasn't in it — backgrounded long enough (or killed by iOS), the workout ended with a score built from the few minutes the app was awake for. The band had the whole thing at 1 Hz anyway, so I rescore sessions from that once it syncs, on read and after each sync. Old workouts fix themselves. Share cards no longer print `0.0` for a session that was never scored.

**Steps (#183).** With phone steps on, the today tile and steps screen added the band's live count on top of the phone's count of the same walk. The toggle also didn't re-derive when switched on, so with the band disconnected it granted permission and kept showing a dash. And a live workout showed "0 STEPS" whenever the band's 100 Hz stream wasn't up — a dash now, since nothing was measured.

**Both imports failing with a FormatException (#199, #160).** A WHOOP export is a zip of CSVs and a `.noopbak` is a zip around noop's sqlite db, and I was feeding both straight into a UTF-8 decoder — hence "Unexpected extension byte (at offset 10)". WHOOP zips are now unpacked and imported, a `.noopbak` gets a message pointing at the raw sensor CSV, and a file I can't read explains why. An export I can read but don't recognise (e.g. downloaded in another language) now fails with the columns it found instead of reporting "imported 0 days".

**Background battery drain on Android (#200).** I requested `CONNECTION_PRIORITY_HIGH` once at connect and never lowered it, so an ~11 ms connection interval was held around the clock on a link that's meant to stay up permanently. It now tracks what's happening: high during a sync or a live workout, balanced when idle, low power in the background. Separately, the 10-minute backfill went through the un-rate-limited "manual" path and so ignored the 15-minute floor while backgrounded; it leaves that to the engine's floored timer now. Band battery is polled every 5 minutes rather than every 30 seconds.

**Band never reconnects after being off the wrist or out of range (#208).** The retry loop only started on a connected → disconnected transition, and the whole loop was inside one try/catch — so one throw anywhere in it ended reconnect for the life of the process, which on Android is forever because the foreground service keeps the process alive. Attempts are contained individually now, and a supervisor checks every minute that a reconnect is actually being attempted. The bond-refusal pause had the same shape (only cleared inside the bond-success path that the pause prevents from running) and now expires after 30 minutes.

Not fixed here: reading a `.noopbak`'s database. The importer points at the raw sensor CSV instead.

## Summary by CodeRabbit

- **New Features**
  - Import WHOOP and NOOP exports from ZIP archives with clearer format errors.
  - Automatically reconcile and rescore recent workouts using available heart-rate data.
  - Added recovery for stalled reconnections and temporary bond-refusal pauses.
  - Improved Bluetooth power management and battery polling efficiency.

- **Bug Fixes**
  - Prevented phone and wearable steps from being double-counted.
  - Preserved unavailable step and strain values instead of displaying misleading zeros.
  - Improved health-platform activity mapping and malformed CSV handling.
  - Improved reconnect reliability across background and lifecycle transitions.


abdulsaheel:
@coderabbit review

abdulsaheel:
@coderabbit review

abdulsaheel:
@coderabbitai review

abdulsaheel:
Fair, and it's the limit I noted in the test itself. Not doing it here though.

`BleEngine` builds its own `BluetoothDevice.fromId` internally, so there's no seam to inject a fake through — the existing engine tests only reach paths where flutter_blue_plus throws on the host. Adding one is a transport refactor of a ~3k-line file that owns the offload and the ACK invariants, and I'd rather not do that in a bug-fix PR.

What is covered: the flag's effect on the target and `sendInit` clearing it (checked that the test fails if the `_connectSetup` term is removed). What isn't: the assignment inside `_doConnect`. Left the comment in the test saying so.

PR #211, opened 2026-08-08
Pin the link-priority decision point so #200 cannot come back

svssathvik7:
> **Rewritten.** This PR originally carried its own implementation of the connection-interval fix. #209 had already landed one — better — an hour before I opened this, off a `main` I had not fetched. The conflict is resolved by adopting #209 wholesale and reducing this to the one thing it does not have. See the comment below for the full account.

## What this adds

`link_priority_policy_test.dart` pins the stepping **rule**. Nothing pins that the engine still goes **through** it.

Issue #200 was not a wrong policy — it was no policy at all: a literal

```dart
device.requestConnectionPriority(connectionPriorityRequest: ConnectionPriority.high)
```

sat in the connect path, ran once, and nothing ever stepped it back down. A future refactor of that path can re-add exactly that line, and every behavioural test in the repo stays green, because `desiredLinkPriority` still returns the right answer — to a caller that no longer exists.

There is no BLE fake here to assert against, so this greps, in the same spirit as `no_debug_only_apis_test.dart` (and for the same reason: the failure is invisible to any test that runs the code).

## The four assertions

1. Exactly **one** real `requestConnectionPriority` call in `lib/`, and it is in the engine.
2. That call is inside `_applyLinkPriority` **and** takes its target from `desiredLinkPriority` — so a request cannot bypass the policy via `ConnectionPriority.values[...]` or any other selector.
3. Every `ConnectionPriority.<value>` literal lives inside that method's mapping switch.
4. Exactly **one** `desiredLinkPriority` declaration, in `sync/sync_policy.dart`.

(4) is load-bearing because (3) is scoped per-file: a second copy of the policy elsewhere in `lib/` would carry its own literals and quietly satisfy (3) while the engine stopped being the single decision point.

Comments **and** string literals are stripped before matching. Both matter — this file names the offending API in prose, and the engine's own failure path logs `requestConnectionPriority(...) failed`, so a naive text count reports two calls where there is one. Method bodies are located by brace depth rather than indentation, so reformatting won't silently vacate the check.

## Verification

Mutation-tested rather than assumed:

| Mutation | Result |
|---|---|
| re-add the original literal request at connect | fails (1) — two call sites — and (3) — literal outside the mapper |
| add a second `desiredLinkPriority` declaration in another `lib/` file | fails (4) |

On the CI-pinned Flutter 3.41.6: `flutter analyze lib test` clean; full suite `+1484 ~2`, 0 failures (bare `main` is `+1480 ~2`).

Test-only change. No production code, no `kAlgoVersion` bump.

<sub>One unrelated observation while running the suite: `workout_reliability_test.dart` — "a queued job stays parked for the session, then runs on release" failed once under full-suite concurrency (`Expected: <1> Actual: <0>`) and passed 3/3 in isolation and in two subsequent full runs. Looks like a timing flake, not something this touches. Flagging in case it is already known.</sub>

## Summary by CodeRabbit

* **Improvements**
  * Bluetooth connection priorities are now mapped consistently, including correct low-power behavior during background idle operation.
  * Priority handling is centralized to provide predictable connection behavior.

* **Tests**
  * Added coverage for all supported priority mappings and background policy behavior.
  * Added safeguards against duplicate or inconsistent priority decisions.
  * Expanded source-analysis regression coverage for reliable validation.

### DIFF
--- a/lib/ble/ble_engine.dart
+++ b/lib/ble/ble_engine.dart
@@ -641,6 +641,22 @@ class BleEngine {
         hasLiveConsumer: _liveEnabled && !_liveHrOnly,
       );
 
+  /// The last hop: the policy's [LinkPriority] as the radio's own enum.
+  ///
+  /// Lifted out of [_applyLinkPriority] because inline it was the one step
+  /// nothing covered. `desiredLinkPriority` could keep returning exactly the
+  /// right answer while every arm here mapped to `ConnectionPriority.high` —
+  /// which IS issue #200, the link pinned at ~11.25 ms overnight — and the
+  /// whole suite stayed green. Arm-by-arm coverage is in
+  /// `link_priority_policy_test.dart`.
+  @visibleForTesting
+  static ConnectionPriority connectionPriorityFor(LinkPriority want) =>
+      switch (want) {
+        LinkPriority.high => ConnectionPriority.high,
+        LinkPriority.balanced => ConnectionPriority.balanced,
+        LinkPriority.lowPower => ConnectionPriority.lowPower,
+      };
+
   @visibleForTesting
   void debugBeginConnectSetup() => _connectSetup = true;
 
@@ -685,11 +701,7 @@ class BleEngine {
         final generation = _linkGeneration;
         try {
           await session.device.requestConnectionPriority(
-            connectionPriorityRequest: switch (want) {
-              LinkPriority.high => ConnectionPriority.high,
-              LinkPriority.balanced => ConnectionPriority.balanced,
-              LinkPriority.lowPower => ConnectionPriority.lowPower,
-            },
+            connectionPriorityRequest: connectionPriorityFor(want),
           );
           // Only remember it if the link we asked is still the live one. A
           // teardown during the await clears `_appliedPriority` precisely so
--- a/test/dart_source_test.dart
+++ b/test/dart_source_test.dart
@@ -1,0 +1,105 @@
+// Regression cases for the scanner the structural tests are built on.
+//
+// Every case here is a way the previous per-line regex helper got it wrong.
+// They matter because when this primitive fails, it fails invisibly: the grep
+// test it feeds goes green either way.
+
+import 'package:flutter_test/flutter_test.dart';
+
+import 'support/dart_source.dart';
+
+void main() {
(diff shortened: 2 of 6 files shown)


abdulsaheel:
@svssathvik7 can you fix the merge conflict

svssathvik7:
@abdulsaheel Conflict resolved, but not the way a conflict fix usually goes — flagging it explicitly rather than burying it in a merge.

The conflict was not textual drift. #209 and this PR were two complete implementations of the same fix in the same region of `ble_engine.dart`, so "resolving" it meant picking one. I branched off a `main` I had not fetched; #209 had merged 54 minutes before I opened this, and I did not notice until reviewing CodeRabbit's comments.

**#209's implementation is better than the one I had here, on every axis I was unsure about:**

- the policy lives in `sync/sync_policy.dart` — where CONTRIBUTING says it belongs; mine was in the engine
- `LinkPriority` keeps the policy file free of `flutter_blue_plus`; mine used `ConnectionPriority` directly
- `_priorityInFlight` / `_priorityRestale` serializes concurrent applies and recomputes the target inside the loop; mine had an unguarded check-then-set
- `_linkGeneration` + `identical(_session, session)` rejects a reply that lands after teardown; **mine wrote the dead session's value back**, which would make the next connection skip its own request and silently sit at Android's default — defeating the whole fix on the drain path
- `_connectSetup` folded into `offloadActive` gets the interval up before INIT; mine raced it

So I dropped mine entirely and reset onto `main`. What is left is the one thing #209 does not have: a structural guard that the engine keeps routing through `desiredLinkPriority`. Test-only, +175/−0.

CodeRabbit's review here was accurate — three of its five findings were real bugs in my implementation, and #209 already avoids all three. Its fourth was that my guard counted raw text occurrences; the version in this PR strips comments and string literals, matches call syntax rather than the bare name, locates bodies by brace depth, and requires the sole call to take its argument from `desiredLinkPriority`. It is mutation-tested against both re-adding the original literal and duplicating the policy.

If you would rather not carry another grep-style test, close it — no argument from me. The value is narrow and specific: #200 shipped because a call site drifted away from a policy that still looked correct in isolation, and nothing in the suite could see that.

PR #212, opened 2026-08-08
fix the crashes coming in from crashlytics

abdulsaheel:
### **User description**
went through what crashlytics has been collecting since 0.9.14 and fixed what is
still live in 0.9.24.

opening Journey could throw and leave the timeline chart blank. the activity
bands clamped their right edge using the left edge as the lower bound, and the
time projection saturates at the plot width, so a workout that started in the
last pixel of the day inverted the clamp. that takes out the whole chart, not
just the one band, and it repeats on every repaint — which is why the event
count is so much higher than the number of people hitting it. the geometry is
its own function now so the edge cases are pinned. five other charts had the
same shape of clamp where the upper bound can fall below the lower one on a
narrow layout; those are fixed too, along with a coach hypnogram that could
build a negative-width rect from a model-authored spec.

sharing a data export from the history screen died on iOS 26 with
"sharePositionOrigin: argument must be set". the share sheet is a popover on
iPhone there now, not just on iPad. the other share buttons already passed an
anchor but fell back to null when the render box was missing — null is the same
crash, and so is a zero rect, so they all go through one helper that returns a
real one.

respiratory rate on the day curve re-ran its estimate once per heartbeat instead
of once every five minutes. the cadence cursor only advanced when the estimate
came back present, and daytime readings are movement-confounded so absent is the
normal case, not the exception. on a noisy day that burns minutes of cpu inside
the isolate that has ninety seconds to finish the second half of a day's derive.
that timeout is the most widely hit issue on the list and this is why. the
all-day hrv curve had the same shape, plus a window sum that ran before its own
cadence gate.

worse, a day that lost its second half then overwrote the complete version of
itself. the write replaces the whole row, and re-deriving deliberately revisits
days that were already finished, so a timeout on a re-run destroyed naps,
workouts, recovery and curves that had been computed correctly days earlier.
that one is silent — no crash, the day just gets thinner. it carries the previous
detail forward now, keyed on absence rather than null so a value that was
genuinely measured as absent stays absent.

a headless start paced its first derive as if the app were on screen. the
background flag defaults to foreground and otherwise only moves on a transition,
so nothing ever set it for a wake that begins in the background. the same gap
was fixed for the ble link a while back, one line above.

map tiles that fail with no network are no longer reported as app errors, and
uncaught network failures generally are no longer filed as crashes. a dropped
tile fetch was counting against crash-free users and burying the real ones. the
tile provider is also rebuilt when the map remounts, because the layer closes
its http client on dispose and an empty-then-refilled route came back with a
dead one.

algo version goes to 60: the two curve fixes change what lands in the stored
bundle.

a few things i looked at and left alone. the anr cluster is not the derive
pipeline starving the main thread — the substrate paging is split across page
reads that each yield, and the two hangs that really were main-isolate blocks
are already fixed. there is real waste there (one day's substrate is loaded
three times over overlapping windows) but it is a performance issue, not the
cause of the anrs, and i would rather not claim a fix i cannot demonstrate. the
readiness_absent non-fatals are all from builds before 0.9.19 and cannot be
emitted by current code. the platform-channel oom and the trig anrs are fixed in
what is already shipping.

still open: making the per-day deadlines survive ios suspending the app, and
giving the scheduler a way to re-pace work that is already in flight. both need
more thought than a patch.

___

### **PR Type**
Bug fix, Enhancement

___

### **Description**
- Fixes derivation timeouts and data loss.
  - Advances HRV/RSA cadence cursor on failed attempts.
  - Prevents partial derivations from overwriting rich details.
  - **Analytics output changed**: Bumped `kAlgoVersion` to 60.
  - Changes absent data handling for daytime RSA/HRV.

- Fixes iOS 26 share sheet crashes.
  - Provides safe, non-zero anchor rect fallbacks.

- Fixes UI rendering crashes.
  - Prevents inverted `clamp` bounds in charts.
  - Fixes negative-width rects in hypnograms.

- Improves crash reporting accuracy.
  - Classifies transient network errors as non-fatal.
  - Silences map tile network exceptions.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  Derive["Derivation Engine"] -- "Second half fails" --> CheckPrev["Check Previous Result"]
  CheckPrev -- "Has rich details" --> Merge["Carry Forward Details"]
  Merge -- "Save" --> DB["LocalDb"]
  Derive -- "Noisy data (absent)" --> Advance["Advance Cadence Cursor"]
  Advance -- "Prevents CPU burn" --> DB
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><details><summary>12 files</summary><table>
<tr>
  <td><strong>derivation_engine.dart</strong><dd><code>Fix derivation timeouts and prevent data loss</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-0e401d8cdb1091d9a9591e9d92c4c5e979293f8480da203b22011a7cf9072ea6">+98/-7</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>app_state.dart</strong><dd><code>Fix headless start pacing</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-d56ba858e512e18dc3962da198adee46d354aca39ff7c1b4a11d5f19b9afc1ae">+4/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>workout_share_card.dart</strong><dd><code>Use safe share origin for workouts</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-3218b777fdaf336c75ef80b845d616ac0f180b8f1459ac6d06799f131c0c945d">+2/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>coach_render.dart</strong><dd><code>Fix negative-width rects and inverted clamps</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-d5ea03a7884dc5697177d8b6dcb06e15ed9c64f911838c59b15aaab948fb6657">+5/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>coach_settings_screen.dart</strong><dd><code>Add mounted guards after async gaps</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-ad3b47e7c1266a4b2fff29798c1cd2d4291606f0b88101e2ff5e2b52fdf86495">+2/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>charts.dart</strong><dd><code>Fix inverted clamp bounds in tooltips</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-ae6fc6ab57ac709cf91c572e6d1095f2331dc288675901583a16a2ce0dbdc486">+4/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>route_map.dart</strong><dd><code>Fix map tile disposal and silence network errors</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-f60a7519d587bfc9a7a577bad4b03dbf544ca19fe579683f3b67e8bc48ab47cf">+22/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>pairing_screen.dart</strong><dd><code>Add mounted guard after bluetooth check</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-e2d6b9cc45b3b1fd7608140e11303deb1059c3a34d4eb14402a9e58ae92de933">+3/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>data_history_screen.dart</strong><dd><code>Use safe share origin for data exports</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-9547ba1d3477d6fbf9800a7076088060739f1faac547daf15885756028fad299">+15/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>profile_screen.dart</strong><dd><code>Use safe share origin for profile exports</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-84850f115ee94849e171ca073ceabf7528d37007bea1de48e67fe0649d96e849">+2/-4</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>recap_screen.dart</strong><dd><code>Use safe share origin for recap sharing</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-352ed161b4e53b21a3a22fc34c7ce6442c06efbd64ca1717781a07edb9d474a3">+3/-5</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>timeline_screen.dart</strong><dd><code>Fix inverted clamp bounds in timeline charts</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-054da88b7ad24d8a94dc27932dba1d31ef69d93fa943788be676c8400149617d">+34/-5</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Error handling</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>error_classification.dart</strong><dd><code>Add transient network error classification</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-e569ed010978d1a28c1bb6080046822b1836c87d585a851793ecc8f67203a2f2">+39/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>telemetry_service.dart</strong><dd><code>Report transient errors as non-fatal</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-1377388a2f920a4e598fa88333d0520d928c2dd5e181bed8c8a3b5e1e884b796">+10/-1</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Miscellaneous</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>kit.dart</strong><dd><code>Export share_origin helper</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-7786eb415d6eaf752f566b4e8ea809d1afbabe3ec946de129545409262fe4426">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
<tr>
  <td><strong>share_origin.dart</strong><dd><code>Add safe share origin helper for iOS 26</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-62d433ba07834d5c6e4e081c2ec1c989de237fc0d2e6da1c6cd827d59ea082ab">+22/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>crash_regressions_test.dart</strong><dd><code>Add regression tests for fixed crashes</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/212/files#diff-46c6e5ba285d593a97d23956dea65f972e8523faa51f4b088b12b82e16e89af1">+176/-0</a>&nbsp; </td>

</tr>

</details>

___

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved share-sheet positioning across database, profile, recap, and workout exports.
  - Prevented crashes and visual glitches in narrow charts, activity timelines, sleep graphs, and range markers.
  - Improved map behavior when routes are empty or map tiles fail to load.
  - Prevented screens from updating after being closed during pairing, settings, and export actions.
  - Improved health metric processing when data is incomplete or updated in stages.
  - Improved handling and reporting of temporary network failures.

### DIFF
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -597,7 +597,18 @@ import 'substrate.dart';
 //   persisted bundle does, so days derived at v58 must be re-derived to pick it
 //   up. `ABSENT` was deliberately NOT invented as a fifth tier — `Tier.all` in
 //   analytics is a closed set of four published grades.
-const int kAlgoVersion = 59;
+//
+// v60 - the all-day HRV and respiratory curves advance their cadence cursor on
+//   every ATTEMPT rather than only on a successful estimate. `_dayRespCurve`
+//   left `lastEmit` unset whenever rsaRespRate came back absent — and absent is
+//   the EXPECTED daytime case, because daytime RSA is movement-confounded — so a
+//   confounded stretch re-ran the triple Lomb-Scargle once per beat instead of
+//   once per five minutes. That is what exhausted the 90 s day-blocks budget and
+//   left days persisted headline-only. `_dayHrvCurve` had the same shape plus an
+//   O(window) sum that ran before its cadence gate was checked. Both curves keep
+//   their sampling intent; points that were previously emitted a beat or two
+//   after a failed attempt now land on the next cadence tick instead.
+const int kAlgoVersion = 60;
 
 // Fold idempotency, the minimum-nights warm-up, and legacy-payload handling
 // all live in SleepProfilePolicy (pure, unit-tested) — see
@@ -2508,6 +2519,35 @@ class DerivationEngine {
     final finalized =
         !producedNothing && (forceFinalize || (ageFinalized && secondHalfOk));
 
+    // A failed/timed-out second half yields a headline-only bundle, and
+    // putDayResult replaces the row wholesale — so re-deriving an already
+    // complete day (rescanRecent deliberately revisits finalized days) destroyed
+    // its naps, sleep periods, workouts, HRR, wear and curves. Carry the
+    // previous result's detail forward instead of blanking it. Same principle as
+    // the producedNothing guard above and the skip-marker guard in
+    // _markDaySkipped: never let a thinner result overwrite a richer one.
+    var effectiveFinalized = finalized;
+    var effectivePartial = !secondHalfOk;
+    if (!secondHalfOk) {
+      final existing = await LocalDb.dayResult(day.date);
+      if (_isRealDayResult(existing)) {
+        final prev = _decodeBundle(existing!['payload_json']);
+        if (prev != null) {
+          final recovered = carryForwardDetail(prev, bundle);
+          final prevPartial = (existing['partial'] as num?)?.toInt() == 1;
+          if (recovered && !prevPartial) {
+            // The day is now as complete as it was before this pass, so it is no
+            // longer partial and may keep the finalized flag it had earned.
+            effectivePartial = false;
+            effectiveFinalized = finalized ||
+                (existing['finalized'] as num?)?.toInt() == 1;
+          }
+          _log('derive ${day.date}: second half failed — carried the previous '
+              'result\'s detail blocks forward (partial=$effectivePartial)');
+        }
+      }
+    }
+
     final scalars =
         (bundle['scalars'] as Map?)?.cast<String, dynamic>() ?? const {};
     double? sc(String k) => (scalars[k] as num?)?.toDouble();
@@ -2518,8 +2558,8 @@ class DerivationEngine {
       windowJson: jsonEncode(
         ((day.sleepJson['window'] as Map?) ?? const {}).cast<String, dynamic>(),
       ),
-      finalized: finalized,
-      partial: !secondHalfOk,
+      finalized: effectiveFinalized,
+      partial: effectivePartial,
       rhr: sc('rhr'),
       rmssd: sc('rmssd'),
       readiness: sc('readiness'),
@@ -2720,6 +2760,46 @@ class DerivationEngine {
     final scalars = payload['scalars'];
     if (scalars is Map && scalars.values.any((v) => v != null)) return true;
     return row['rhr'] != null || row['rmssd'] != null || row['readiness'] != null;
+  }
+
+  /// Fill [next]'s missing detail from [prev] when the second-half compute
+  /// failed, so a headline-only pass never blanks a day that already had naps,
+  /// workouts, HRR, wear and curves. Returns true if anything was carried over.
+  ///
+  /// Keyed on ABSENCE, not on null: isolate 1 writes its headline scalars
+  /// explicitly and a null there is a real "we could not measure this today"
+  /// that must survive. Only keys the failed second half never got to add are
+  /// restored — a freshly computed value always wins.
+  @visibleForTesting
+  static bool carryForwardDetail(
+    Map<String, dynamic> prev,
+    Map<String, dynamic> next,
+  ) {
+    var carried = false;
+    for (final e in prev.entries) {
+      if (e.key == 'scalars' || e.key == 'series') continue;
+      if (next.containsKey(e.key) || e.value == null) continue;
+      next[e.key] = e.value;
+      carried = true;
+    }
+    // `scalars` and `series` are flat maps the second half patches INTO rather
+    // than owning, so they merge per key instead of wholesale.
+    for (final sub in const ['scalars', 'series']) {
+      final p = prev[sub];
+      if (p is! Map) continue;
+      final n = next[sub];
+      if (n is! Map) {
+        next[sub] = Map<String, dynamic>.from(p.cast<String, dynamic>());
+        carried = true;
+        continue;
+      }
+      for (final e in p.entries) {
+        if (n.containsKey(e.key) || e.value == null) continue;
+        n[e.key] = e.value;
+        carried = true;
+      }
+    }
+    return carried;
   }
 
   /// Test seam for [_markDaySkipped] — the "a skip marker must never destroy a
@@ -3897,7 +3977,10 @@ class DerivationEngine {
       while (ts[i] - ts[lo] > winMs) {
         lo++;
       }
-      if (i - lo >= 10) {
+      // Cadence gate FIRST: the sum-of-squared-differences below is O(window),
+      // and running it for every beat only to discard the result on the 60 s
+      // check was the whole window's work wasted per sample.
+      if (i - lo >= 10 && ts[i] - lastEmit > 60000) {
         var ssd = 0.0;
         var nd = 0;
         for (var k = lo + 1; k <= i; k++) {
@@ -3909,14 +3992,17 @@ class DerivationEngine {
           ssd += d * d;
           nd++;
         }
-        if (nd >= 8 && ts[i] - lastEmit > 60000) {
+        if (nd >= 8) {
+          // Advance on a computed estimate, not only on one that passed the
+          // plausibility filter — otherwise a stretch of artifact-heavy beats
+          // re-runs the window every beat.
+          lastEmit = ts[i];
           final rmssd = math.sqrt(ssd / nd);
           if (rmssd <= 220) {
             out.add({
               't': (ts[i] / 1000).round(),
               'v': double.parse(rmssd.toStringAsFixed(1)),
             });
-            lastEmit = ts[i];
           }
         }
       }
@@ -3953,12 +4039,17 @@ class DerivationEngine {
         final nnt = [for (var k = lo; k <= i; k++) ts[k] - t0];
         final est = ana.rsaRespRate(nn, nnt, artifactFraction: 0.15);
         final brpm = est.present ? est.value!.brpm : null;
+        // Advance the cadence cursor on every ATTEMPT, not just on a successful
+        // estimate. Daytime RSA is movement-confounded (see above), so absent is
+        // the common case — and while lastEmit sat inside the success branch a
+        // confounded stretch re-ran the triple Lomb-Scargle once per BEAT
+        // instead of once per 5 min. That is what blew the day-blocks budget.
+        lastEmit = ts[i];
         if (brpm != null) {
           out.add({
             't': (ts[i] / 1000).round(),
             'v': double.parse(brpm.toStringAsFixed(1)),
           });
-          lastEmit = ts[i];
         }
       }
     }
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -2534,16 +2534,19 @@ class DerivationEngine {
         final prev = _decodeBundle(existing!['payload_json']);
         if (prev != null) {
           final recovered = carryForwardDetail(prev, bundle);
-          final prevPartial = (existing['partial'] as num?)?.toInt() == 1;
-          if (recovered && !prevPartial) {
-            // The day is now as complete as it was before this pass, so it is no
-            // longer partial and may keep the finalized flag it had earned.
-            effectivePartial = false;
-            effectiveFinalized = finalized ||
-                (existing['finalized'] as num?)?.toInt() == 1;
-          }
+          final prevVersion = (existing['algo_version'] as num?)?.toInt();
+          final outcome = recoveryOutcome(
+            recovered: recovered,
+            prevPartial: (existing['partial'] as num?)?.toInt() == 1,
+            prevVersion: prevVersion,
+            prevFinalized: (existing['finalized'] as num?)?.toInt() == 1,
+            finalizedByAge: finalized,
+          );
+          effectivePartial = outcome.partial;
+          effectiveFinalized = outcome.finalized;
           _log('derive ${day.date}: second half failed — carried the previous '
-              'result\'s detail blocks forward (partial=$effectivePartial)');
+              "result's detail blocks forward (v$prevVersion -> v$kAlgoVersion, "
+              'partial=$effectivePartial)');
         }
       }
     }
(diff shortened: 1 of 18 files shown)


abdulsaheel:
all four were real, thanks. fixed in 6a1acb8.

the engine one is the important one and it means my original fix did not do what
i claimed. `background` is final on `DerivationEngine` and selects both the
concurrency and the per-day timeout, so seeding `_deriveScheduler` left the
engine itself on the foreground budget for a headless first sweep — i fixed the
pacing gate and left the thing it was supposed to protect untouched. it is
constructed with the initial value now. `_derive` is late-initialized and
`_background` is assigned in both constructors well before anything reads it, so
the value is correct at construction without reordering.

the hrv cursor is out of the `nd >= 8` check. i had underestimated this: the
window is five minutes, so a few hundred beats, and a stretch too noisy to yield
eight clean pairs redid that entire sum on every subsequent beat.

timeouts are out of the type set. a derivation, database or lifecycle timeout
raises the same class as a network one, and those are exactly the ones worth
keeping fatal.

both share sheets check mounted after their export or capture.

on the cadence tests — i added them but they are weaker than they look and i
would rather say so than let them stand as more than they are. they pin the
spacing invariant, that an unusable stretch stays empty and terminates, and that
a clean stretch after a bad one still produces points. that last one is the real
guard, because it covers the risk this change introduces: advancing on failure
could have silenced the curve once quality returned.

what they do not cover is the per-beat path itself. i tried to build a failing
test for it and could not — i could not construct synthetic rr that makes
`rsaRespRate` return absent, so every input i tried took the success branch and
ran identically before and after the change (76ms both ways, then 100ms both
ways on broadband jitter). the pathological behaviour needs real confounded
daytime rr, and short of a captured fixture i do not have a deterministic way to
provoke it. so the cost fix is argued from the code path, not demonstrated by a
test, and i would rather that be explicit.

abdulsaheel:
good call, that was the right answer to the gap i described. done in 8a815c9.

both curves now carry a seam: an override for the respiratory estimator and an
attempt counter on each. the counter is the part that matters — the fix is about
how often the work runs, not what it returns, so attempts are the only thing
that separates the fixed code from the broken code. output assertions cannot see
it, which is why my earlier tests passed either way.

forcing the estimator to abstain across a two hour stretch:

- respiration attempts 7970 times before the fix, 26 after
- the hrv window sum runs 5990 times before, 102 after

i checked both by reverting each cursor move in turn and confirming the test
fails, rather than only that it passes. respiration fails on `Expected: a value
less than or equal to <26> Actual: <7970>`, hrv on `<102>` against `<5990>`.

that also puts a number on the original claim, which i had only argued from the
code path: on a confounded stretch the respiratory estimator was running about
300 times more often than intended, and each of those runs is a triple
lomb-scargle over a three minute window. that is the 90 s budget.

abdulsaheel:
right, fixed in 991a303. workout_share_card read the anchor after the frame
wait, the raster, the png encode and the file write — the one call site that did
not follow the contract the helper documents. it would not have crashed, the
fallback covers a detached box, but the rect could describe a box that had since
moved. all four sites measure before their awaits now.

abdulsaheel:
this one was the best catch of the review — fixed in a9511fe.

it is specific to this pr bumping the algo version, and it would have hit every
device on the first derive after updating. `LocalDb.dayResult` is
`ORDER BY algo_version DESC LIMIT 1` with no filter, so the row offered for
carry-forward is the v59 one. if the second half then failed, my recovery
carried v59 detail into the v60 bundle, wrote it under v60 and inherited v59's
finalized flag — locking last version's resp and hrv curves in under this
version's number. those curves are the entire reason for the bump, and a
finalized row is never revisited, so nothing would ever have recomputed them.
my fix for silent data loss had a path that quietly preserved the wrong data.

a cross-version carry now stays partial and unfinalized so a later pass redoes
it. same-version carries behave as before.

one thing worth flagging: writing this i nearly introduced a regression of my
own. moving the decision into a helper, my first version returned
`finalized: false` on every fallback path, which would have broken the import
case — imports force-finalize even when partial, because there is no stored raw
to recompute them from. the fallback passes `finalizedByAge` through now and
there is a test for it.

the decision is a pure function so the cases are pinned directly, and there is a
persisted-row test in derive_result_protection_test that seeds a v59 row and
asserts the lookup really does return it. i verified both fail with the version
check removed rather than only that they pass.

abdulsaheel:
@coderabbit review

Issue #184, opened 2026-08-04
Strength Workouts Not Writing to Apple Health


ramishroshan:
### What happened?

Openstrap Edge fails to write "Strength" workout data to Apple Health. Other workout types sync and write data to Apple Health without issue. The failure is isolated entirely to the Strength category.

### Steps to reproduce

1. Complete a "Strength" workout in Openstrap Edge.
2. Open Apple Health and navigate to workout data.
3. Observe that the Strength workout from Openstrap Edge is missing, whereas other workout types from the app populate correctly.

### App version

0.9.21

### Platform

iOS (TestFlight)

### Phone model and OS version

14 pro max ios 27 beta 3

### Which band?

WHOOP 4.0

### Screenshots

<img width="645" height="1398" alt="Image" src="https://github.com/user-attachments/assets/6e54fade-d808-41c1-a479-585b52105174" />
<img width="645" height="1398" alt="Image" src="https://github.com/user-attachments/assets/95c26670-ab7b-4797-bd22-30fa06df1952" />

### Quick checks

- [x] The official WHOOP app is not also connected to this band
- [x] I've searched existing issues for this

abdulsaheel:
Fixed in [v0.9.24](https://github.com/OpenStrap/edge/releases/tag/v0.9.24).

HealthKit has no bare `STRENGTH_TRAINING` — it rejected every write, and the error went to a `debugPrint` so nothing surfaced. iOS now sends `TRADITIONAL_STRENGTH_TRAINING`. Swims had the mirror-image bug on Android and were fixed at the same time.

Please reopen if you encounter it again.

PR #217, opened 2026-08-09
more workout types, a filter on the list, and calories that stop guessing

abdulsaheel:
### **User description**
Five people asked for the same three things in Discord and the discussions, so this is those three.

**Types.** The picker had nine, which meant boxing, rowing, hiking, skiing, stair work, pilates, racquet sports and the ball sports all had to be logged as "Other" — and reached Apple Health as an unlabelled workout. There are twenty-one now. A type only earns a tile if a pack actually draws it, so nothing ships with a stand-in icon that looks like something else. Adding a tile without adding the matching health mapping is silent — the workout still exports, just as a generic entry — so a test fails on it now.

That pushed the start sheet past the height a bottom sheet gives itself by default, which is exactly how "Log a past workout" once ended up invisible and untappable. The sheet scrolls now, and the old guard against content falling off the edge became a guard that every type can still be reached.

**Filtering.** The list only ever filtered by timeframe. It now filters by type, by minimum duration and by minimum strain, and sorts by longest or hardest as well as by date. Two details worth naming: a narrowed list recomputes its own summary, so the totals above it always describe what is actually on screen; and a session that is happening right now is never hidden by a floor it has not had time to clear.

**Calories.** The last place the app still made something up. The live tick substituted a 30-year-old 70 kg male for whatever the profile was missing, so an untouched profile produced a confident calorie total — for somebody else's body, and reading high for anyone lighter than the stand-in. The re-score path has always refused to guess. Both share one predicate now, and a session with no anchors shows a dash rather than a number, all the way out to the shareable card.

___

### **PR Type**
Enhancement, Bug fix, Tests

___

### **Description**
- Expanded workout types from 9 to 21, each with distinct glyph and health mapping

- Added workout list filtering by type, duration, strain, and sort order

- Fixed calorie fabrication: unanchored profiles now show "—" instead of a stand-in body's estimate

- Added tests for calorie anchors, filter logic, and health mapping coverage

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Profile.hasCalorieAnchors\n(age + mass + sex)"] -- "gates" --> B["Live tick\ncalorie accrual"]
  A -- "gates" --> C["Substrate re-score\nmanual session"]
  B -- "null when unanchored" --> D["caloriesOrNull\n(LiveWorkoutState)"]
  D -- "dash or count-up" --> E["Finish card / share card\n/ session sheet"]
  F["kWorkoutTypes\n(21 types)"] -- "picker tiles" --> G["workoutTypeSheet\n(scrollable)"]
  F -- "health mapping" --> H["healthActivityForType\n(iOS + Android)"]
  I["WorkoutFilter\n(type/duration/strain/sort)"] -- "apply()" --> J["Filtered list\n+ recomputed summary"]
  J --> K["_FilterBar\n(workouts_screen)"]
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><details><summary>5 files</summary><table>
<tr>
  <td><strong>profile.dart</strong><dd><code>Add hasCalorieAnchors predicate for Keytel formula</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-9aba12ce5a4e98153b52ff855ea5960cfe10b3541e2594d870d5b84735bb1bf1">+13/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>manual_session.dart</strong><dd><code>Gate substrate re-score calories on hasCalorieAnchors</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-4c14c9e94dd086da44e897bce8da931cac7c7096b23884f616ae76b95c14465a">+5/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>app_state.dart</strong><dd><code>Remove fallback defaults from live calorie tick, use caloriesOrNull</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-d56ba858e512e18dc3962da198adee46d354aca39ff7c1b4a11d5f19b9afc1ae">+33/-12</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>live_session_screen.dart</strong><dd><code>Propagate nullable calories through finish card and share</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-a1806a05926c11f526d297b01e0d2448c4db55d6b57cfabc57d26e7e36c7081a">+22/-10</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>workout_share_card.dart</strong><dd><code>Accept nullable calories, dash out on share card</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-3218b777fdaf336c75ef80b845d616ac0f180b8f1459ac6d06799f131c0c945d">+6/-2</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Enhancement</strong></td><td><details><summary>6 files</summary><table>
<tr>
  <td><strong>os_icons.dart</strong><dd><code>Add 12 new sport glyphs for expanded workout types</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-71d8d9053b4c8dca5493b60f416db2e29b347a8d6e43f07b1e50b00c57a3ce28">+46/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>workout_types.dart</strong><dd><code>Expand picker from 9 to 21 types, fix scrollable sheet</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-6304f71690f750cd6a55b8b5a60021cbf9ea41db5c94e4c5572ee93dc31eb11b">+54/-14</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>workout_filter.dart</strong><dd><code>New filter/sort policy for workout list</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-2b382b26bf97ce974a3199eabdc85a96bfe9ab7b25f35deb6366558d3014c274">+166/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>workout_filter_sheet.dart</strong><dd><code>New bottom sheet UI for filter and sort controls</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-20f5754cfcb250bb9a4a5e822e4510511c1369d8d50a1ea8a84df7d81c868bb0">+169/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>workouts_screen.dart</strong><dd><code>Wire filter bar, recompute summary over filtered list</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-d90667f8b853f130ad9d0e1bf052b328a038e5a551b36e3b78a622d895e44cf7">+164/-44</a></td>

</tr>

<tr>
  <td><strong>health_export.dart</strong><dd><code>Map 13 new workout types to platform health activities</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-7688da159b76fc36f532ed153ea690343655e60ab0436820d1bcbc54113e15f5">+39/-0</a>&nbsp; &nbsp; </td>

</tr>
<tr>
  <td><strong>workout_calorie_anchors_test.dart</strong><dd><code>Tests for hasCalorieAnchors and caloriesOrNull contract</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-56cb78b793b2af2b2c89670c411282eb1975d7f4bdcb512e95c949e8eabc4fcc">+108/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>workout_filter_test.dart</strong><dd><code>Unit tests for filter, sort, and summarize logic</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-f94db7577c9ad32c81b1c0353e171e4198b64da6fb8ab7c1e4acc1733f0fb6bc">+209/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>workout_health_mapping_test.dart</strong><dd><code>Extend health mapping tests for all new types</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-0318b9eb433e9b393c281f1dcb12510aa7f7f6c90c4c9c69e57fbc3ba72acfb3">+63/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>workouts_header_actions_test.dart</strong><dd><code>Update sheet test to assert scrollable grid reachability</code>&nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/217/files#diff-709b6535fbafe54e6c660e122b41e79f17246cc9ffd53876e2e0082c5a241e36">+46/-41</a>&nbsp; </td>

</tr>

</details>

___

## Summary by CodeRabbit

* **New Features**
  * Added workout filtering and sorting by type, duration, strain, date, and workout metrics.
  * Added expanded workout categories, including boxing, rowing, hiking, climbing, skiing, snowboarding, stairs, Pilates, tennis, basketball, soccer, and golf.
  * Added scrollable workout-type selection, filter controls, live result counts, and summaries.
  * Added matching activity icons and improved health-platform activity mapping.

* **Bug Fixes**
  * Calorie estimates now remain unavailable when required profile information is missing.
  * Live sessions, summaries, and sharing show unavailable calories as “—” instead of zero.

### DIFF
--- a/lib/compute/manual_session.dart
+++ b/lib/compute/manual_session.dart
@@ -310,7 +310,11 @@ ManualSessionStats computeManualSessionStats({
       strainFromPerMinuteHr(perMin, profile: profile, restingHr: restingHr);
 
   double? calories;
-  if (hrMax != null && age != null && weightKg != null && sex != null) {
+  if (profile.hasCalorieAnchors &&
+      hrMax != null &&
+      age != null &&
+      weightKg != null &&
+      sex != null) {
     // Real anchors only — `usedDefaultAnchors` stays false, so we are never
     // persisting a kcal figure built on a fabricated 220/60.
     final bout = ana.Calories.estimateBoutCalories(
--- a/lib/compute/manual_session.dart
+++ b/lib/compute/manual_session.dart
@@ -513,6 +513,19 @@ ReconciledSessionScore reconcileSessionScore({
   // same minutes, so the larger is the better estimate and the smaller is just
   // a less complete view.
   T? better<T extends num>(T? live, T? sub) {
+    // `sub ?? live`, NOT `sub`. A null from a complete substrate is "I have
+    // nothing to say", not "the answer is nothing": a max HR of null means no
+    // worn samples survived, while the live tally actually watched the session
+    // happen, and a null strain means the profile no longer carries the anchor
+    // the score needs — neither is grounds for destroying a real measurement
+    // taken when it did.
+    //
+    // The cost of that is real and accepted: a calorie figure fabricated by an
+    // older build (30 y / 70 kg / male, before the live tick learned to
+    // abstain) is never cleared by a re-score. Making the null authoritative
+    // would heal those, and would also wipe legitimately scored sessions
+    // whenever the substrate happens not to be able to score them, which is
+    // the worse trade.
     if (substrateIsComplete) return sub ?? live;
     if (live == null) return sub;
     if (sub == null) return live;
--- a/lib/compute/profile.dart
+++ b/lib/compute/profile.dart
@@ -49,4 +49,17 @@ class Profile {
 
   bool get isComplete =>
       ageYears != null && weightKg != null && heightCm != null && sex != null;
+
+  /// The anchors Keytel (2005) needs to turn heart rate into kcal: age, body
+  /// mass and sex. Height is not one of them, so it is deliberately absent
+  /// here — gating calories on [isComplete] would refuse to score a profile
+  /// that has everything the formula actually reads.
+  ///
+  /// The one definition of "can we cost this session in calories", shared by
+  /// the live tick and the substrate re-score. They used to disagree: the
+  /// re-score refused to guess while the live tick silently substituted a
+  /// 30-year-old 70 kg male, so an unfinished profile produced a confident
+  /// kcal number that was simply somebody else's.
+  bool get hasCalorieAnchors =>
+      ageYears != null && weightKg != null && sex != null;
 }
--- a/lib/health/health_export.dart
+++ b/lib/health/health_export.dart
@@ -1118,6 +1118,45 @@ HealthWorkoutActivityType healthActivityForType(
       return HealthWorkoutActivityType.YOGA;
     case 'hiit':
       return HealthWorkoutActivityType.HIGH_INTENSITY_INTERVAL_TRAINING;
+    case 'boxing':
+      return HealthWorkoutActivityType.BOXING;
+    case 'rowing':
+    case 'row':
+      return HealthWorkoutActivityType.ROWING;
+    case 'hike':
+    case 'hiking':
+      return HealthWorkoutActivityType.HIKING;
+    case 'climb':
+    case 'climbing':
+      // Bare `CLIMBING` is iOS-only; `ROCK_CLIMBING` is the one spelling both
+      // stores accept, which is the #184 lesson applied ahead of the bug.
+      return HealthWorkoutActivityType.ROCK_CLIMBING;
+    case 'ski':
+    case 'skiing':
+      // `SKIING` is Android-only; `DOWNHILL_SKIING` exists on both.
+      return HealthWorkoutActivityType.DOWNHILL_SKIING;
+    case 'snowboard':
+    case 'snowboarding':
+      return HealthWorkoutActivityType.SNOWBOARDING;
+    case 'stairs':
+    case 'stair':
+      // `STAIRS` is iOS-only; `STAIR_CLIMBING` exists on both.
+      return HealthWorkoutActivityType.STAIR_CLIMBING;
+    case 'pilates':
+      return HealthWorkoutActivityType.PILATES;
+    case 'tennis':
+    case 'racquet':
+    case 'squash':
+    case 'padel':
+    case 'badminton':
+      return HealthWorkoutActivityType.TENNIS;
+    case 'basketball':
+      return HealthWorkoutActivityType.BASKETBALL;
+    case 'soccer':
+    case 'football':
+      return HealthWorkoutActivityType.SOCCER;
+    case 'golf':
+      return HealthWorkoutActivityType.GOLF;
     default:
       return HealthWorkoutActivityType.OTHER;
   }
--- a/lib/health/health_export.dart
+++ b/lib/health/health_export.dart
@@ -1154,7 +1154,18 @@ HealthWorkoutActivityType healthActivityForType(
       return HealthWorkoutActivityType.BASKETBALL;
     case 'soccer':
     case 'football':
-      return HealthWorkoutActivityType.SOCCER;
+      // `SOCCER` passes the plugin's Dart-side guard but is COMMENTED OUT of
+      // Health Connect's Kotlin write map (HealthPlugin.kt, "TODO: add
+      // soccer"), so the call reaches the channel and comes back
+      // `success(false)` rather than throwing. This file treats a false as a
+      // genuine write failure and counts it toward the day's give-up budget —
+      // so one soccer workout would silently pause that day's ENTIRE export,
+      // resting HR and sleep included. Worse than #184, which at least failed
+      // only itself. OTHER is accepted on Android, so the workout lands
+      // unlabelled instead of taking the day down with it.
+      return ios
+          ? HealthWorkoutActivityType.SOCCER
+          : HealthWorkoutActivityType.OTHER;
     case 'golf':
       return HealthWorkoutActivityType.GOLF;
     default:
--- a/lib/state/app_state.dart
+++ b/lib/state/app_state.dart
@@ -4439,7 +4439,7 @@ class AppState extends ChangeNotifier {
             4.184;
       }
       // Add per-second slice (kcal/min / 60). Clamp to 0 in case of low HR.
-      w.calories += (kcalMin.clamp(0.0, 30.0) / 60.0);
+      w.accrueCalories(kcalMin.clamp(0.0, 30.0) / 60.0);
 
       // Strain is NOT accrued here. `accrueHr` (called above) recomputes it
       // from the session's per-minute HR through the one shared Banister ->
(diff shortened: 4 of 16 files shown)


abdulsaheel:
Went through the three flagged spots.

The summary total was worth changing. Adding `?? 0` and skipping a null come to the same number, so the sum was never wrong — but writing it as a default invites someone to later swap the zero for a "reasonable" estimate, which is exactly the fabrication the rest of the change removes. It skips explicitly now, and the test says what it means: an uncosted session still counts as a session and still contributes its minutes, it just adds nothing to the kcal total.

The mid-session profile change can't happen. `LiveWorkoutState.profile` is final and is only ever set at construction, from the profile the session is started under — that's deliberate, a workout is scored against the profile it was performed under rather than whatever the profile says when it ends. Filling in a profile during a workout leaves the running session alone; the substrate re-score picks it up afterwards and costs the whole window properly, which is the path that was built for exactly this.

The navigator one you withdrew yourself on the second look — agreed, the `mounted` guard is there.

abdulsaheel:
Both of these were right.

The live-session exception was too broad. Type is known the moment a session starts, so filtering to runs should never surface the ride that happens to be in progress — only the duration and strain floors need the exception, because a session two minutes old has not had a chance to clear them yet. Split accordingly, with a test for the live-cycle-under-a-run-filter case.

The case sensitivity was a real gap, and it was in all three lookups rather than just the label one — the icon lookups compared against the raw string too, so a row written as `Strength` would have fallen through to the generic fallback glyph. The stored `type` column is free-form text and older rows and imports do carry mixed case, so all three normalize once now. There is also a test asserting every key in the table is lowercase, since that is the assumption the normalization rests on.

abdulsaheel:
@coderabbitai review

abdulsaheel:
The strain sort one was worth taking. Null strain means the session was never scored — no profile anchor, or no heart rate for the window — and ranking that as a genuine zero puts "we could not score this" in among "you barely moved". It sorts below every scored session now, including real zeros.

The same question applies to the filter, and there the answer goes the other way: a strain floor excludes an unscored session, because it cannot be shown to clear the bar and listing it under "strain 10+" would be a claim there is no evidence for. Both are pinned by tests.

The summary total was already changed in an earlier commit on this branch — it skips a null rather than defaulting it, for exactly the reason described.

abdulsaheel:
Ran a second pass over this myself, since CodeRabbit's last run came back rate limited rather than actually reviewed. It found one thing that should not have shipped.

`soccer` mapped to `HealthWorkoutActivityType.SOCCER`, which is in the health plugin's Dart-side list but **commented out of Health Connect's Kotlin write map**. That combination is worse than issue #184 rather than the same as it: #184 threw before the platform channel and failed only itself, whereas this passes the Dart guard, reaches the channel, and comes back `success(false)`. The exporter treats a false as a genuine write failure and counts it toward that day's give-up budget — so a single soccer workout would have silently paused that whole day's export, resting HR and sleep along with it. It maps to `OTHER` on Android now.

The test I added specifically to prevent this class of bug could not have caught it, which is the more useful half of the finding: its Android set was transcribed from the plugin's Dart list, and for Android that is the wrong source of truth. The Dart list is advisory; the Kotlin map is the gate, and here the two disagree. Both sets are now pinned against the file that actually decides, and the header says which one that is per platform.

Two smaller things from the same pass. `hike` now records a GPS route — it sits directly beside `walk` in the picker and picking it signals more outdoor intent, not less, and route points cannot be backfilled. Ski and snowboard deliberately do not: the route screen leads with distance and pace, and pace down a lift-served descent would read as a performance figure while measuring gravity. And `climb` was drawing mountains right next to a person hiking, where the mountains were the tile that read as hiking; it is a ladder now.

abdulsaheel:
Second self-review pass, this one on correctness. Six findings, five taken and one rejected — the rejected one is the interesting one.

**Taken.** The workout detail screen still printed `0 kcal` for a null column, which is the exact failure this PR exists to remove, sitting in the most-viewed place a session appears — and the share card exported from that same screen already read "—". A live session with a complete profile but no heart rate at all (link dropped, strap off) accrued nothing and reported a confident `0`; "can we score this" and "did we score this" are different questions and both have a zero-shaped answer, so calories track whether the estimate ever ran now, and agree with strain, which already called that case absent. The filter pill could overflow its row — "Basketball, Snowboard · 90m+ · strain 17+" is reachable and the header is about 350 pt wide — so it ellipsizes. A filtered summary was dropping the auto-type accuracy note, which describes the whole range and has nothing to do with a filter. And imported sessions carry the export's own spelling, so filtering by Run hid every imported run while the feed showed cards titled "Running" — there is one alias table now, shared by the filter and by every label and icon lookup, so a row cannot read as one type and filter as another.

**Rejected, after trying it.** The observation that a calorie figure fabricated by an older build is never cleared by a re-score is correct. The obvious fix is to let a complete substrate's null be authoritative instead of falling back to the stored value — I made that change, and `session_score_reconcile_test.dart` caught it immediately, which is what it was written for. That fallback is deliberate: a null max HR from a complete substrate means no worn samples survived, while the live tally actually watched the session happen, and a null strain means the profile no longer carries the anchor the score needs. Neither is grounds for destroying a real measurement taken when it could be. Healing the old fabricated values that way would also wipe legitimately scored sessions whenever the substrate happens not to be able to score them, which is the worse trade. Reverted, and the reasoning is now written down at the call site so the next person does not have to rediscover it.

So: sessions logged before this change keep whatever the old tick wrote. Everything logged after it is honest, and no migration deletes history on a guess.

abdulsaheel:
Both taken.

The summary total was the same fabrication one level up: `summarizeWorkouts` started at 0 and stayed 0 when every session in the range was uncosted, so the card would have read "0 kcal" for a timeframe where nothing could be costed at all. It returns null now and the card dashes out, matching what a single session already did.

The two lookup mechanisms could not actually have diverged — `kWorkoutTypesByKey` is a comprehension over `kWorkoutTypes`, so adding a type to one adds it to the other by construction. But having one helper walk the list while its sibling used the map was pointless inconsistency, so all four now go through the map.

abdulsaheel:
@coderabbitai review

PR #218, opened 2026-08-09
journal entries that carry a number, not just a tag

abdulsaheel:
### **User description**
The most-requested journal changes — mood, water, caffeine with a dose and a time, a sleep-quality rating, custom entries — all need the same thing the journal did not have: a number. A tag can only say whether something happened, and one coffee versus five is usually the actual question.

Nine built-in fields ship: mood, sleep quality, energy, stress, soreness, water, caffeine, alcohol, screens before bed. Beyond those you can define your own — name it, say whether it is a 1–5 rating, an amount, or minutes, and it behaves exactly like a built-in everywhere downstream. Caffeine and alcohol also ask when the last one was, because a 200 mg morning coffee and a 200 mg evening one are the same dose and a completely different night, and that is the half that predicts anything.

**Absent and zero are different answers, and every control here is shaped around it.** "No caffeine today" is a measurement; "I didn't fill this in" is not. Reading the second as the first invents a data point at the very bottom of the dose range, which is exactly where a correlation is most sensitive. So: a field starts unset, one tap on minus reaches a real zero rather than making you step down to it, any value can be cleared back to unset, and clearing one deletes the row instead of leaving a stale reading nobody made. Saving a day replaces that day outright rather than merging into it, for the same reason.

Ratings are 1–5, not 1–10 — a ten-point self-report is not ten distinguishable states, and the extra resolution is noise a rank correlation then has to see through.

Deleting a custom field forgets its label, not its history. Those readings were still real, and they render unlabelled until the field is defined again.

Schema goes to 28. Two new tables, purely additive — an upgrade touches no existing journal row, and a day that only ever had tags keeps them and simply has no numeric rows. There is a migration test from v27 pinning that.

Correlating these fields is deliberately **not** in this PR. That needs a rank statistic that does not exist in the analytics package yet (OpenStrap/analytics#41), and holding storage and UI hostage to that merge would just make both harder to review. Values are recorded and shown here; the insight wiring follows once the pin moves.

One drive-by: the journal screen kept its own private copy of the preset tag list, which is how it would have drifted from the compose screen the first time either changed.

___

### **PR Type**
Enhancement, tests

___

### **Description**
- Adds numeric journal fields (mood, sleep quality, energy, stress, soreness, water, caffeine, alcohol, screens) with absent-vs-zero semantics

- Schema bumped to v28: two new tables (`journal_metric`, `journal_field_def`), purely additive; existing tag/note rows untouched

[...]

PR #219, opened 2026-08-09
blood work, and an export a spreadsheet can open

abdulsaheel:
### **User description**
Two asks from the discussions that turned out to share a shape — both are about data the band cannot produce.

**Labs.** Twenty-five common markers across blood count, iron, metabolic, lipids, hormones, vitamins, inflammation and organ panels, each with a fixed unit and typical adult reference intervals; anything missing you can add yourself.

The reference intervals are the part that can do harm, so the constraints are deliberate. They exist because a bare "ferritin 42" is a spreadsheet with extra steps — but they are context, never a verdict. Every laboratory publishes its own interval, derived from its own assay and population, and they differ enough that a value inside one lab's range sits outside another's, so **the range on your own report is the one that applies to you**. A value outside ours reads as "outside the typical range", not "high" or "abnormal", and nothing anywhere suggests what to do about it. Several markers genuinely differ by sex and carry both intervals; with no sex in the profile they show **no interval at all** rather than defaulting to one, because guessing would flag a large share of perfectly normal results. A custom marker gets no range invented for it.

Two storage decisions worth calling out. Each result stores the unit it was entered under rather than looking it up, so a reading can never be silently reinterpreted if the catalogue's canonical unit changes — turning 400 ng/mL into 400 nmol/L would be the worst fabrication available here. And a result is keyed on (marker, date drawn), so re-entering one corrects a typo instead of stacking a second reading that a trend would then average.

**CSV export**, beside the existing `.db` export. Opening your data in a spreadsheet and restoring it onto a new phone are different jobs, and a SQLite file only does the second. Six files — daily metrics, workouts, sleep stages, metric history, journal, labs — all read through the same derived views the coach reads, so an export can never contain something the app would not show you and can never reach raw sensor rows or GPS. A metric that was never computed is written as an **empty field, never a 0**: the difference is unrecoverable once it is in a spreadsheet, and a column of zeroes is exactly the kind of thing someone averages. UTF-8 BOM so Excel on Windows does not mangle notes.

Schema goes to 29, purely additive. Note it assumes #218 (schema 28) lands first — if that one is going to wait, tell me and I will rebase this to 28.

___

### **PR Type**
Enhancement, tests

___

### **Description**
- Adds hand-entered blood-work (Labs) with 25 built-in markers, sex-aware reference intervals, and custom marker support

- Adds CSV export (6 files: daily metrics, workouts, sleep stages, metric history, journal, labs) via the profile screen

- Bumps SQLite schema from v27 to v29 with new `lab_result` and `lab_marker_def` tables; no `kAlgoVersion` bump (no analytics output changed)

- Adds tests for CSV escaping/absence semantics, catalogue invariants, and lab result storage idempotence

___

### Diagram Walkthrough

```mermaid
flowchart LR
  catalogue["lab_catalogue.dart\n25 built-in markers\nsex-aware ref ranges"]
  db["db.dart\nschema v29\nlab_result + lab_marker_def tables"]
  entrySheet["lab_entry_sheet.dart\nAdd / edit / delete result"]
  labsScreen["labs_screen.dart\nMarker cards with range context"]
  csvExport["csv_export.dart\n6 CSV sets via coach views"]
  profileScreen["profile_screen.dart\nLabs + CSV export entries"]
  tests["test/\ncsv_export_test\nlab_catalogue_test\nlab_result_store_test"]

  catalogue -- "marker definitions" --> entrySheet
  catalogue -- "range lookup" --> labsScreen
  db -- "CRUD" --> entrySheet
  db -- "CRUD" --> labsScreen
  entrySheet -- "modal sheet" --> labsScreen
  labsScreen -- "pushed from" --> profileScreen
  csvExport -- "rawQuery via coach views" --> profileScreen
  db -- "schema migration" --> db
  tests -- "covers" --> csvExport
  tests -- "covers" --> catalogue
  tests -- "covers" --> db
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>6 files</summary><table>
<tr>
  <td><strong>lab_catalogue.dart</strong><dd><code>New: 25 built-in blood markers with sex-aware reference intervals</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/219/files#diff-5dc6467c75a2bebb1064980848ba31878720fbb295816fb246a96b8c29597a8f">+394/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>db.dart</strong><dd><code>Schema v29: add lab_result and lab_marker_def tables with CRUD</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/219/files#diff-b96f416000a07a912224053a34f440c27993e014d324d6ae8c9dfe88bb0617e6">+120/-1</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>csv_export.dart</strong><dd><code>New: RFC 4180 CSV export for 6 data sets via coach views</code>&nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/219/files#diff-143affd8d235b28909750b02682f2e6ded53ede62ce750ce54800b64a29ef31b">+190/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>labs_screen.dart</strong><dd><code>New Labs screen: marker cards with range context and history</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/219/files#diff-ff357e49203d88e30c960a74dee15f08731a7be618fa493b4d2cca85c6b9da14">+315/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>lab_entry_sheet.dart</strong><dd><code>New bottom sheet: add, edit, or delete a lab result</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/219/files#diff-3121a4eac44b4b0d8e4f36097122da53124bdb1d7c6d8f88b86e64430206bb8f">+374/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>profile_screen.dart</strong><dd><code>Add Labs and CSV export entries to profile screen</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/219/files#diff-84850f115ee94849e171ca073ceabf7528d37007bea1de48e67fe0649d96e849">+47/-0</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>csv_export_test.dart</strong><dd><code>Tests: CSV escaping, null-as-empty, and schema query coverage</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/219/files#diff-755ab4be62f1b1d3dfc958235548c725dc90d7ba2b3a4026acdd6617f097cb91">+150/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>lab_catalogue_test.dart</strong><dd><code>Tests: catalogue invariants, range selection, and custom keys</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/219/files#diff-1d2867cb554b75d6b05818544e0b829422fdcfd11caa82835dc4985a56b5da52">+180/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>lab_result_store_test.dart</strong><dd><code>Tests: lab result upsert idempotence and unit preservation</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/219/files#diff-e0724821f185bc78bfc68f813dd89886b84f2d58b0d19e1d349c1dfaa54c501c">+180/-0</a>&nbsp; </td>

</tr>

</details>

___

## Summary by CodeRabbit

- **New Features**
  - Added a Labs section for recording, viewing, editing, and deleting laboratory results.
  - Added built-in and custom lab markers with units, categories, notes, and sex-specific reference ranges.
  - Added CSV export for metrics, workouts, sleep, journal entries, lab results, and history.
  - Added sharing of generated CSV files, with feedback for empty or partial exports.

- **Bug Fixes**
  - Preserved existing lab readings when marker definitions are deleted.
  - Improved CSV handling for special characters, empty values, and spreadsheet formulas.

### DIFF
--- a/lib/data/csv_export.dart
+++ b/lib/data/csv_export.dart
@@ -122,7 +122,17 @@ const kCsvExportSets = <CsvExportSet>[
   ),
 ];
 
-/// RFC 4180 field escaping.
+/// Characters that make Excel, Google Sheets and LibreOffice treat a cell as a
+/// FORMULA rather than text.
+///
+/// Journal notes, journal tags and lab notes are free text the user typed, and
+/// these files are handed to a share sheet — so whoever opens the spreadsheet
+/// executes whatever a cell starting with one of these evaluates to. A note
+/// beginning "=" is a formula in every mainstream spreadsheet, and formulas
+/// can reach the network and the filesystem.
+final _formulaLeaders = RegExp(r'^[=+\-@\t\r]');
+
+/// RFC 4180 field escaping, plus formula-injection neutralisation.
 ///
 /// Null becomes an EMPTY field rather than the string "null" or a 0 — the
 /// distinction between "not measured" and "measured as nothing" has to survive
@@ -134,9 +144,13 @@ String csvField(Object? v) {
       // false impression of precision the metric does not have.
       ? (v == v.roundToDouble() ? v.toInt().toString() : v.toString())
       : v.toString();
+  // A leading apostrophe is the convention every mainstream spreadsheet reads
+  // as "this is text" — it is not displayed, and the value stays legible.
+  // Applied only to strings: a negative NUMBER starts with `-` and must stay a
+  // number, or every negative delta in the file becomes unusable text.
+  if (v is String && _formulaLeaders.hasMatch(s)) s = "'$s";
   if (s.contains(RegExp('[",\n\r]'))) {
-    s = '"${s.replaceAll('"', '""')}"';
-    return s;
+    return '"${s.replaceAll('"', '""')}"';
   }
   return s;
 }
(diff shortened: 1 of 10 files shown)


abdulsaheel:
All thirteen were real. The formula injection one is the one I am glad was caught — these files are handed straight to a share sheet, so a journal note beginning with `=` runs on whoever opens the spreadsheet, and mainstream spreadsheets will happily let a formula reach the network. Text cells are neutralised now. Numbers deliberately are not: a negative delta has to stay a number or the numeric columns become unusable.

The temp-file accumulation was the same class of problem in slow motion — plaintext readiness, sleep, journal notes and blood work, one more copy per export, nothing ever removing them. One export's worth exists at a time now.

The return contract was genuinely broken and I had not noticed: every set wrote a header-only file even with zero rows, so the caller's "nothing to export" branch was only reachable when *everything* failed, which meant a total failure was reported to the user as an empty database. Empty sets write nothing, failures come back by name, and the two messages are now different sentences.

On the day deletion question — the exemption is intentional and the docstring says so now. "Delete this day" in the data manager is about reclaiming space from sensor data; a blood panel is neither sensor data nor large, it was typed in by hand on a different screen, and it has its own delete there. Losing a year-old result because the band data from that date was cleared would be a surprising deletion, not a tidy one.

The unit-authority finding was the most useful of the batch. The storage layer was already right — every row carries the unit it was entered under — but both UI sites read the unit from the resolved marker instead, which quietly undid that. Editing a row now keeps its own unit, and a reading whose custom marker was deleted stays fully readable *and* editable rather than becoming a number nobody can correct.

Schema is 28 rather than 29 now, so the ladder has no hole if this lands before #218.

abdulsaheel:
Both real, and the first one was mine — the previous round's fix for files piling up introduced it. Clearing every earlier run at the start of a new export is fine right up until you notice that the export returns before the share sheet has finished reading, so a second export deletes the files the first share is still handing over. Each run has its own directory now and the previous one survives, which bounds the copies on disk without letting a new export destroy an open share. The row also refuses to re-enter while a share is in flight, since that await covers exactly the dangerous window.

Worth noting the test I wrote for the earlier fix was asserting the wrong thing — that the previous run was gone — so it would have locked the bug in. It now asserts the opposite, alongside a second test that the retention is still bounded.

The two mounted guards are in.

PR #221, opened 2026-08-09
correlations that read numbers, blood work in your profile, backups, naps you can correct, and breathing you can do with your eyes shut

abdulsaheel:
### **User description**
Five of the six remaining items from the discussions triage. Per-app vibration patterns is parked.

**Journal correlations.** The typed fields shipped with nowhere to go — recorded, displayed, correlated with nothing. This moves the analytics pin to pick up `journalNumericCorrelations` and wires it in. Findings lead with a slope in the outcome's own units ("about 4 ms less HRV per extra coffee") rather than a coefficient, with rho as supporting detail. Deliberately a separate pass from the tag correlations rather than more rows in the same list: a tag answers "were those days different", a dose answers "does more of this go with less of that", and they carry different evidence — a difference of means with a Cohen's d versus a rank correlation with a confidence interval. Its date axis is the days a *number* was recorded, which is not the same set as the days a tag was.

**Profile from the health store.** Weight, height, and on Apple sex and date of birth. Weight and height are adopted because they drift and three things the app computes depend on them; age and sex only fill a gap, because neither changes and another app's record should not overrule something you set deliberately. Sex and DOB are requested **only on Apple** — Health Connect has no characteristic record for either, and requesting them there is issue #184's shape again. Android's manifest gains `READ_WEIGHT`/`READ_HEIGHT` with no write counterparts; without them Health Connect silently returns nothing, which is indistinguishable from an empty store.

**Automatic backup.** Daily or weekly, into an `OpenStrap Backups` folder in the app's Documents directory, reachable from Files (`UIFileSharingEnabled` + `LSSupportsOpeningDocumentsInPlace`) and from the Android file manager. Point iCloud Drive, Synology or Nextcloud at it. Not a user-picked folder: that needs a persisted SAF tree URI or a security-scoped bookmark, both of which expire silently, and a backup that quietly stopped working is worse than one that lives somewhere slightly less convenient. It runs **on foreground when due** rather than pretending to a background schedule neither platform will honour, and the sheet says so — along with the fact that it is not encrypted.

**Naps you can correct.** Three people asked, from two directions: "it tracked sleep when I was awake, let me delete it" and "I napped for two hours and it wasn't counted". Same feature — the detector's answer is a proposal and the person who was there gets the last word.

Two decisions worth calling out. Edits are stored **separately** from the detector's output and replayed over it on every derivation, rather than written into the result: the detector improves, and a day re-derived under a better stager should still respect "there was no nap here" instead of freezing the old detection alongside the edit. And a rejection matches by **overlap, not exact bounds** — detected boundaries shift between runs, and an edit that stopped applying when one moved by a minute would let a deleted nap quietly come back.

**This bumps `kAlgoVersion` to 61**, and that is the point rather than a side effect: logged naps count toward sleep totals, sleep debt and readiness exactly as detected ones do. Days carrying an edit are force-derived alongside sleep-override days so an edit to a finalized day takes effect.

**Breathing.** The screen already had real cardiac coherence from live RR; it was missing the other three things. Four patterns instead of one fixed pace. A session that ends when the button said it would — it used to say "2-Minute Session" and run until you stopped it. And the strap buzzes each phase change, which was always the point of an exercise you are not supposed to be watching a phone during. Sessions are kept now, so the score is a trend rather than a number that vanished on stop. Coherence stays on resonance only: it measures oscillation at the paced frequency, and scoring box breathing against it grades it on an exam it is not sitting.

**Interval timer** rides the same phase engine — rounds and breaths are the same problem. It records nothing and starts no session; a workout already covers that, and a timer that quietly created sessions would double-count every round. Reachable from the start-workout sheet, which is only safe because that sheet became scrollable earlier.

Schema 29 → 31 across three additive rungs, with a migration test that walks v27 all the way up and writes to every new table — a CREATE with a typo still leaves a table nothing can write to.

___

### **PR Type**
Enhancement, Bug fix, Tests

___

### **Description**
- Five new user-facing features: journal numeric correlations, health-store profile import, automatic backup, correctable naps, and improved guided breathing

- `kAlgoVersion` bumped 60→61 and `schemaVersion` bumped 29→31; two new tables (`sleep_nap`, `breathing_session`) added with migrations

- Breathing screen gains four patterns, real session timer, per-phase strap buzz, and session history; new interval timer screen shares the same phase engine

- Nap edits (add/reject) stored separately from detector output and replayed on every derivation; rejection matches by overlap, not exact bounds

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Health store\n(HealthKit / Health Connect)"]
  B["HealthProfileImporter\n(health_profile_import.dart)"]
  C["AppState.updateProfile()"]
  D["Local profile map"]

  E["User nap edit\n(add / reject)"]
  F["LocalDb.putNapEdit()\nsleep_nap table (schema v31)"]
  G["DerivationEngine\napplyNapEdits()"]
  H["day_result\n(kAlgoVersion 61)"]

  I["Breathing session\n(CalmBreathingScreen)"]
  J["LocalDb.putBreathingSession()\nbreathing_session table (schema v30)"]
  K["AppState.buzzBreathPhase()"]
  L["BLE strap buzz"]

  M["Auto backup\n(auto_backup.dart)"]
  N["Documents/OpenStrap Backups/\n*.db files"]

  A -- "read" --> B
  B -- "mergeHealthProfile()" --> C
  C --> D

  E --> F
  F -- "napEditDays() force-derive" --> G
  G --> H

  I -- "≥60 s session" --> J
  I -- "phase boundary" --> K
  K --> L

  M -- "daily / weekly" --> N
```

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>11 files</summary><table>
<tr>
  <td><strong>derivation_engine.dart</strong><dd><code>Bump kAlgoVersion 60→61; replay nap edits over detector output</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-0e401d8cdb1091d9a9591e9d92c4c5e979293f8480da203b22011a7cf9072ea6">+78/-27</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>db.dart</strong><dd><code>Schema v29→31; add sleep_nap and breathing_session tables</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-b96f416000a07a912224053a34f440c27993e014d324d6ae8c9dfe88bb0617e6">+144/-1</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>nap_edits.dart</strong><dd><code>Pure nap-edit merge logic: applyNapEdits, overlap rejection</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-cf5850a430ae4891c87ceb470713bac74591e46915094eaf78d7f37fbb9bc9cd">+129/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>health_profile_import.dart</strong><dd><code>Read weight/height/sex/DOB from platform health store</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-01e15aa2836a7c0e900f39122198e27edcc34006059c212b15ecbb63c8f8ec3e">+226/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>auto_backup.dart</strong><dd><code>Automatic daily/weekly database backup to Documents folder</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-d9029ccf67209672ed6680bd901dda4332c50ef9229ca7ea5609b900906940df">+176/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>breath_phases.dart</strong><dd><code>Shared phase engine for breathing and interval timer</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-984ece097de2a444d42ee42d6833ece8e08b03ff5681cfbf987776ef43a79c4a">+200/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>calm_breathing_screen.dart</strong><dd><code>Four patterns, real timer, per-phase strap buzz, session history</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-61d24055867b669abe16153614b9c9cda679c1033e9b5bb6e83055862ca4622f">+261/-113</a></td>

</tr>

<tr>
  <td><strong>interval_timer_screen.dart</strong><dd><code>New interval timer screen using shared breath phase engine</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-75fa57559e556cf44b41e62af91067852bc452dc848b9b41d744a3383c4465a6">+253/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>sleep_periods_screen.dart</strong><dd><code>Add/remove nap UI; detected naps suppressed by overlap match</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-33d275e303e3215b5712e4db789fe3d72ea6c502b4c7850e58269fffb8ca9d31">+156/-3</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>profile_screen.dart</strong><dd><code>Health-store import row and automatic backup row in profile</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-84850f115ee94849e171ca073ceabf7528d37007bea1de48e67fe0649d96e849">+121/-1</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>app_state.dart</strong><dd><code>Wire backup, nap-edit reanalysis, breathing pattern/history/buzz</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-d56ba858e512e18dc3962da198adee46d354aca39ff7c1b4a11d5f19b9afc1ae">+131/-3</a>&nbsp; </td>

</tr>
<tr>
  <td><strong>nap_edits_test.dart</strong><dd><code>Tests for nap edit merge, overlap rejection, and validation</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-72c9e3444c2e575cfc26ae945521fd29b0d02404e7e6d026d8abaede9a1eeb00">+209/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>auto_backup_test.dart</strong><dd><code>Tests for backup scheduling, filename ordering, and pruning</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-244ad77994f0f00d444ba941e7512baa468d37c29ba27fb286e4865bd44a5aec">+228/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>health_profile_import_test.dart</strong><dd><code>Tests for health-store snapshot, merge policy, and permissions</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-66dcbd241645a6089221045cae0ba68114380d783afe92ca433d354b3a9cee4f">+239/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>breath_phases_test.dart</strong><dd><code>Tests for phase engine timing and interval pattern helpers</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-9eadf40afc212aad760d9a96a8aa59e36e28a4e81c64320f412a6c8117a5974a">+183/-0</a>&nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Configuration changes</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>prefs.dart</strong><dd><code>Add backup cadence and last-run timestamp preference keys</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-4d226eb2ac5e998bb863d72ff5060e4279461d3e6144245db2fcf70a8a9ea457">+4/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>AndroidManifest.xml</strong><dd><code>Add READ_WEIGHT and READ_HEIGHT Health Connect permissions</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-63272c98a6330850888e633b43ba4c9decee6431a115e0d2731564838eaa1d1d">+7/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>Info.plist</strong><dd><code>Add UIFileSharingEnabled and LSSupportsOpeningDocumentsInPlace</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-13a4d9d8e1fcf0ee1bd48f8728d2b8fbb885f9cca317e047a515d06552458b9f">+9/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Dependencies</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>pubspec.yaml</strong><dd><code>Update analytics sibling pin for journal numeric correlations</code></dd></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-8b7e9df87668ffa6a04b32e1769a33434999e54ae081c52e5d943c541d4c0d25">+10/-1</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Additional files</strong></td><td><details><summary>7 files</summary><table>
<tr>
  <td><strong>app.dart</strong></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-1fb04a3749d56cc322a1bc7e7e6b8988600930c339b0d8df7119837e1b7ae20d">+5/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>local_repository_impl.dart</strong></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-da6b8e6c3188825ab7c7d364093ccefbd9e40a284244154fdbffd372214f6352">+108/-3</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>journal_screen.dart</strong></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-f9fb6a722773613a3a73b813fa3261a83d08a47ecdab6d3596c7213059b024d2">+95/-3</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>workout_types.dart</strong></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-6304f71690f750cd6a55b8b5a60021cbf9ea41db5c94e4c5572ee93dc31eb11b">+17/-2</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>workouts_screen.dart</strong></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-d90667f8b853f130ad9d0e1bf052b328a038e5a551b36e3b78a622d895e44cf7">+31/-1</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>calm_breathing_view_test.dart</strong></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-69c57ff15bfc7338dd92adb45425a545b7bf3a51c6e69f3a9be80f7fa4f1fd24">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>db_migration_ladder_test.dart</strong></td>
  <td><a href="https://github.com/OpenStrap/edge/pull/221/files#diff-10df492a6cb4a0545549f5a609174a97238612bec5863325490b5705a6f035e9">+47/-0</a>&nbsp; &nbsp; </td>

</tr>

</details>

___

## Summary by CodeRabbit

- **New Features**
  - Added manual nap logging, editing, removal, and sleep re-analysis.
  - Added scheduled or on-demand local database backups with retention.
  - Added health-store profile import for weight, height, age, and sex.
  - Added selectable guided-breathing patterns, session history, phase cues, and an interval timer.
  - Added numeric journal insights and correlation summaries.
  - Enabled iOS Files access and in-place document opening.
  - Added Health Connect access for reading weight and height.
- **Bug Fixes**
  - Due backups now run when the app returns to the foreground.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -64,6 +64,13 @@
     <uses-permission android:name="android.permission.health.WRITE_SLEEP" />
     <uses-permission android:name="android.permission.health.READ_EXERCISE" />
     <uses-permission android:name="android.permission.health.WRITE_EXERCISE" />
+    <!-- READ only, and no WRITE counterpart on purpose: body metrics are read
+         INTO the local profile (weight drifts, and the calorie estimate
+         depends on it) and this app has no business writing them back. Health
+         Connect silently returns nothing for an undeclared permission rather
+         than failing, so an omission here looks exactly like an empty store. -->
+    <uses-permission android:name="android.permission.health.READ_WEIGHT" />
+    <uses-permission android:name="android.permission.health.READ_HEIGHT" />
     <uses-feature android:name="android.hardware.bluetooth_le" android:required="true" />
     <!-- CompanionDeviceManager availability (not required: devices without it just skip
          the association and rely on the sticky service + watchdog). -->
--- a/ios/Runner/Info.plist
+++ b/ios/Runner/Info.plist
@@ -61,6 +61,15 @@
 	</array>
 	<key>NSBluetoothAlwaysUsageDescription</key>
 	<string>OpenStrap connects to your WHOOP band over Bluetooth to sync your health data.</string>
+	<!-- The automatic-backup folder lives in the app's Documents directory.
+	     Without BOTH of these it exists but is invisible in Files, so nothing
+	     the user might sync it with (iCloud Drive, Nextcloud, a cable) can
+	     reach it — which makes the whole feature pointless rather than broken,
+	     and therefore easy to miss. -->
+	<key>UIFileSharingEnabled</key>
+	<true/>
+	<key>LSSupportsOpeningDocumentsInPlace</key>
+	<true/>
 	<key>NSHealthShareUsageDescription</key>
 	<string>OpenStrap reads your step count from Apple Health to show your daily steps, and reads back its own recent samples so it never writes duplicates.</string>
 	<key>NSHealthUpdateUsageDescription</key>
--- a/lib/app.dart
+++ b/lib/app.dart
@@ -88,6 +88,11 @@ class _OpenStrapAppState extends State<OpenStrapApp> with WidgetsBindingObserver
       // already-running process (openAppWhenRun doesn't guarantee a fresh
       // launch) — the constructor-time check alone would miss that case.
       unawaited(app.checkPendingSiriRoute());
+      // Backups run on foreground, when due — there is no background scheduler
+      // that works on both platforms, and a schedule that claims "daily" while
+      // delivering whenever the OS feels like it is worse than one that is
+      // honest about when it fires.
+      unawaited(app.runBackupIfDue());
       if (app.isPaired) app.openSession();
     } else if (state == AppLifecycleState.paused) {
       // Backgrounded: hand the band to the iOS restore path so it can wake-and-drain
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -4365,6 +4365,54 @@ class DerivationEngine {
   /// distinguishable only by HOW the abstention happened. A reader that checks
   /// `bundle['naps']?['value'] == null` and one that checks
   /// `bundle.containsKey('naps')` would disagree.
+  /// Abstention path that still honours what the USER logged.
+  ///
+  /// The detector abstains on exactly the days this feature exists for — strap
+  /// off for part of the afternoon, a short record, a failure. Returning early
+  /// there dropped every logged nap on the floor: no card to see, no minutes
+  /// credited, and no way to delete the row the user had just created, while
+  /// the edit kept force-re-deriving that day forever.
+  ///
+  /// A logged nap needs nothing from the detector — it carries its own absolute
+  /// bounds — so it is published on its own. The day is still reported as
+  /// unjudged when the user logged nothing, because that is what it is.
+  static List<Map<String, dynamic>>? _napsWhenUnjudged(
+    Map<String, dynamic> bundle,
+    Map<String, dynamic>? scMap,
+    List<NapEdit> napEdits,
+    String note,
+  ) {
+    final merged = applyNapEdits(const [], napEdits);
+    if (merged.isEmpty) {
+      _writeUnknownNaps(bundle, note);
+      return null;
+    }
+    bundle['naps'] = <String, dynamic>{
+      'value': merged,
+      'count': merged.length,
+      // No detection confidence, because there was no detection. These are
+      // reported, not estimated.
+      'confidence': null,
+      'tier': 'reported',
+      'inputs_used': const ['user'],
+      'note': '$note — showing what you logged',
+    };
+    scMap?['nap_min'] = napMinutes(merged).toDouble();
+    return [
+      for (final nap in merged)
+        {
+          'is_main': false,
+          'onset_ts': nap['start'],
+          'wake_ts': nap['end'],
+          'duration_min': nap['duration_min'],
+          'in_bed_min': nap['in_bed_min'],
+          'efficiency': null,
+          'confidence': null,
+          if (nap['source'] != null) 'source': nap['source'],
+        },
+    ];
+  }
+
   static void _writeUnknownNaps(
     Map<String, dynamic> bundle,
     String note,
@@ -4396,8 +4444,12 @@ class DerivationEngine {
     try {
       final n = s.length;
       if (n < 60) {
-        _writeUnknownNaps(bundle, 'too little 1 Hz data to assess naps');
-        return null;
+        return _napsWhenUnjudged(
+          bundle,
+          scMap,
+          napEdits,
+          'too little 1 Hz data to assess naps',
+        );
       }
       final accel = <ana.AccelSample>[
         for (var i = 0; i < n; i++)
@@ -4423,15 +4475,12 @@ class DerivationEngine {
       );
 
       if (!m.present) {
-        bundle['naps'] = <String, dynamic>{
-          'value': null,
-          'count': null,
-          'confidence': 0,
-          'tier': m.tier,
-          'inputs_used': m.inputs_used,
-          'note': m.note,
-        };
-        return null;
+        return _napsWhenUnjudged(
+          bundle,
+          scMap,
+          napEdits,
+          m.note ?? 'naps could not be assessed for this day',
+        );
       }
 
       final t0 = s.tsSec.first;
@@ -4517,8 +4566,12 @@ class DerivationEngine {
       ];
     } catch (e) {
       if (kDebugMode) debugPrint('[derive] naps FAILED/skipped: $e');
-      _writeUnknownNaps(bundle, 'nap detection failed for this day');
-      return null;
+      return _napsWhenUnjudged(
+        bundle,
+        scMap,
+        napEdits,
+        'nap detection failed for this day',
+      );
     }
   }
 
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -26,6 +26,7 @@ import 'dart:isolate';
 import 'dart:math' as math;
 
 import 'package:flutter/foundation.dart';
+import 'nap_edits.dart';
 import 'package:openstrap_analytics/onehz.dart' as ana;
 import 'package:firebase_core/firebase_core.dart';
 import 'package:firebase_performance/firebase_performance.dart';
@@ -608,7 +609,21 @@ import 'substrate.dart';
 //   O(window) sum that ran before its cadence gate was checked. Both curves keep
 //   their sampling intent; points that were previously emitted a beat or two
 //   after a failed attempt now land on the next cadence tick instead.
-const int kAlgoVersion = 60;
+// v61 - NAP EDITS. The nap detector's answer is now a PROPOSAL: a nap the user
+//   logged is added, and one they rejected is suppressed, replayed over the
+//   detector's output on every derivation rather than written into it (so a
+//   better detector later still respects "there was no nap here"). Rejection
+//   matches by OVERLAP, not by exact bounds, because the detector's boundaries
+//   shift between runs and an edit that stopped applying when a boundary moved
+//   by a minute would be worse than useless.
+//
+//   This moves numbers, which is why it is a version bump rather than a read
+//   path: `nap_min` is summed over the merged list, so a logged nap credits
+//   against sleep need and sleep debt exactly as a detected one does — that
+//   was the explicit product decision, not an accident of where the code sat.
+//   Days carrying an edit are force-derived alongside sleep-override days, so
+//   an edit to an already-finalized day actually takes effect.
+const int kAlgoVersion = 61;
 
 // Fold idempotency, the minimum-nights warm-up, and legacy-payload handling
 // all live in SleepProfilePolicy (pure, unit-tested) — see
@@ -1046,7 +1061,11 @@ class DerivationEngine {
       // FINALIZED (locked) day — it's the user's word. Force those back into the
       // todo set. (No-raw days are guarded in the per-day loop so we never
       // clobber a good manual result with an empty re-derive once raw is pruned.)
-      final overrideDays = await LocalDb.sleepOverrideDays();
+      final overrideDays = {
+        ...await LocalDb.sleepOverrideDays(),
+        // A nap edit on a finalized day has to take effect too — same reason.
+        ...await LocalDb.napEditDays(),
+      };
       final todoDays = [
         for (final day in scope.targetDays)
           if (!finalized.contains(day) || overrideDays.contains(day)) day,
@@ -2416,9 +2435,22 @@ class DerivationEngine {
 
       // Built on THIS isolate so the Isolate.run closure captures only this plain
       // sendable object (never `this`, `day`, or `bundle`).
+      // Read HERE, on the main isolate — the worker has no database.
+      final napEdits = [
+        for (final row in await LocalDb.napEdits(day.date))
+          NapEdit(
+            kind: row['source'] == 'rejected'
+                ? NapEditKind.rejected
+                : NapEditKind.added,
+            startSec: (row['start_ts'] as num).toInt(),
+            endSec: (row['end_ts'] as num).toInt(),
+          ),
+      ];
+
       final blocksInput = _DayBlocksInput(
         daySub: daySub,
         napSub: day.napSub,
+        napEdits: napEdits,
         sleepSub: sleepSub,
         profile: profile,
         onsetSec: day.sleepOnsetSec,
@@ -4357,6 +4389,9 @@ class DerivationEngine {
     int? attributionEndSec,
     List<List<int>> wristOff = const [],
     List<List<int>> charging = const [],
+    // Read on the main isolate and carried in, like every other DB-sourced
+    // input here — this runs inside the compute worker, which has no database.
+    List<NapEdit> napEdits = const [],
   }) {
     try {
       final n = s.length;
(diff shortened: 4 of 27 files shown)


abdulsaheel:
Nine of the twelve taken, and two of them were things that would have shipped broken.

**The Android backup folder was unreachable.** `getApplicationDocumentsDirectory()` resolves to app-private storage there, so backups were being written somewhere no file manager and no sync app could open — the feature would have worked perfectly and been pointless, which is the worst kind of bug because nothing looks wrong. It writes to app-specific external storage now, and the move handles temp and external being different filesystems, where rename fails outright.

**The finite session called stop on every frame.** `onStop` is asynchronous, so between the timer expiring and the parent flipping `active` false, every single frame fired another stop — sixty a second, each one banking a session row. That is mine, and the two-minute session I had just added is exactly what would have triggered it. Latched, with the local clock stopped immediately.

Also taken: retention matched `openstrap-*.db` and would have deleted a user's own `openstrap-notes.db` from the same folder; filenames now carry seconds because two runs in one minute shared a name and the second silently overwrote the first; a backup cannot start while one is running; Health Connect's 30-day cap without the history permission (which `health` 11.1.1 has no API to request, so the honest fix is to ask for what we can actually have rather than declare a permission we cannot use); the wake lock leaked when the interval screen was popped mid-round; the calm screen kept a copy of the resonance pattern that could drift from the table and shipped an empty description; and the journal methodology note still said tagged days only.

**One declined.** The cadence check uses a 24-hour interval rather than local calendar-day boundaries. The guideline about not assuming 86400-second days is about day *labels*, where it is exactly right — but this is a "how often", not a date, and an interval is the better model for it: a backup at 23:00 followed by one at 00:30 should not count as two days. On a DST day the effect is that a daily backup runs an hour earlier or later, never that one is skipped, because the trigger is foreground rather than a fixed alarm. Using calendar days here would introduce the skip it is meant to prevent.

abdulsaheel:
Right, and I had half-fixed it. The guard covered the export but ended before the caller persisted the timestamp, so a second trigger could still read the old value, judge a backup due, and start another one.

The lock covers the whole sequence now — read, decide, export, prune, record. The part that makes it actually work is that `lastRun` and `markRun` are callbacks rather than values: passing the timestamp in would mean the caller had already read it before queueing, which is the same stale read wearing a different hat. The regression test fires two triggers without awaiting the first and asserts the second read sees the first run rather than a stale null, and that exactly one of them decided it was due.

Same-second destinations are allocated against what is already on disk, with a test that two runs at an identical timestamp produce two files rather than one overwritten. The queue also survives a failed run, which it needed to — a chained tail without an error guard would have wedged every later backup for the life of the process after one throw.

abdulsaheel:
@coderabbit review

abdulsaheel:
All three, and two of them are mine from the last round.

The cadence was the same bug I had just fixed for the timestamp, left in place one line above it. Read before the call entered the queue, so a call waiting behind a running export acted on the setting as it was when it queued — someone switching backup off mid-export would still have got another unencrypted copy of their health data written after disabling it. It re-reads inside the lock now, and the test flips the setting from inside the first export to prove the queued call sees it.

The unique-filename search returning its last candidate when all hundred were taken was straightforwardly backwards: it hands the next backup a real snapshot to overwrite, which is the one thing the unique naming exists to prevent. It errors instead, and the test fills every slot and checks all hundred files are still intact afterwards.

And you are right that the queue-recovery test proved nothing — both calls were cadence `off`, so neither reached `_runBackup` and neither could fail. That is worse than no test, because it reads as coverage. The exporter is injectable now, so a failing export is reachable, and the test asserts the backup queued behind it still runs.

abdulsaheel:
@coderabbitai review

abdulsaheel:
CodeRabbit stayed rate limited, so I ran my own adversarial pass over the diff instead. Eight real defects, two of which would have made the features look broken rather than subtly wrong.

**A nap logged on a day the detector abstained on vanished.** All three abstention paths — too little data, detector not present, exception — returned before the edits were applied. So on exactly the days this feature exists for (strap off part of the afternoon, short record) the nap produced no card, credited no minutes, and could not even be deleted afterwards, while the stored edit kept force-re-deriving that day forever. Logged naps carry their own absolute bounds and need nothing from the detector, so they are published on their own now; the day is only reported unjudged when nothing was logged.

**Re-entering the breathing screen mid-session left it dead.** Swipe-back and system back never reach `onBack`, which is the only thing that stops a session — so the session keeps running and the next view mounts with `active` already true, with no state change for `didUpdateWidget` to catch. The clock never started: frozen circle, no haptics, and a timed session that never ended. That is worse than the pre-existing version of the same shape, because the whole session lifecycle now hangs off that clock. Fixed in `initState`, with a regression test.

Also fixed: a logged nap now supersedes a detection it overlaps (the entry-time overlap check only sees a snapshot, so a later re-derive could find the same bout and count the afternoon twice); main sleep is in that overlap check, or a nap could be logged inside the night and counted in both; backgrounding a session banked wall-clock time, so two minutes backgrounded and resumed forty minutes later stored a forty-minute session with a coherence score drawn mostly from unpaced breathing; the interval timer's "three buzzes" for session-end are played by the firmware as one and were indistinguishable from a round change, so it has its own pattern; restoring a backup dropped every sleep correction, and since the edits are the only copy that silently reinstated deleted naps; and the Health import searched ten years back while adopting weight unconditionally, letting an ancient reading overwrite a current profile and then feed calories and BMR.

Two things noted and deliberately not changed: both ticker screens `setState` every frame, rebuilding more than they need to while holding a wake lock — real, but a refactor rather than a fix, and the running-state trees are small. And the backup runs on the same foreground frame as the BLE session re-establishing, which on a large database is contention at a bad moment; worth moving, but not in this PR.

1727 tests.

abdulsaheel:
All five, and the deadline one is a defect in the fix from my last round — fourth consecutive round where that has happened.

Starting the clock on mount fixed the frozen re-entry and introduced a worse version of the same bug: the clock restarted from zero and the length reverted to the picker's default, so a five-minute session re-entered at 4:00 showed 2:00 and stopped almost immediately. The real problem was that the view owned the timing at all — it is remounted every time someone leaves the screen and comes back. Elapsed and remaining now come from when the session actually began and what it was actually asked for, both held by the session. Regression test asserts a 5-minute session remounted at 4:00 shows 0:59 remaining and does not stop.

The tier was a straightforward mistake on my part — `'reported'` is not in the vocabulary, which is a closed set of four. `AUTH` is the one that means directly measured or definitional, and that is what a self-report is: the user is not estimating that they napped, they are stating it.

The backup exporting before choosing a destination is a good catch on a path I had just added a test for without asserting the cleanup. Destination first now, so a run that cannot find a free name never exports at all — and the test asserts the exporter was never called rather than just that the outcome failed.

Also fixed: the `_rederive` provider lookup after two awaited writes, and the Apple window assertion, which passed for any date in 2025 rather than pinning the year.

abdulsaheel:
Taken — and it is the same mistake as the round before, one layer down. I keyed "is there an authoritative target" on the target being non-null, but null is itself a meaningful answer for a running session: it means open-ended. So an open session left and re-entered past two minutes inherited the picker's default, showed a countdown it never had, and stopped on the spot.

The session's `startedAt` is what says whether the session is authoritative; the target is then whatever it is, null included. Test covers a nine-minute open session remounting and continuing to count up.

abdulsaheel:
@coderabbit review

abdulsaheel:
@coderabbitai full review

PR #222, opened 2026-08-09
a 13-week activity heatmap on the workouts screen

svssathvik7:
## What

A 13-week activity heatmap above the workouts feed, shaded by the calories burned in logged workouts.

The session feed is reverse-chronological, so it shows you *what* you did but never the shape of a quarter — where the gaps are, and whether the rhythm is holding. That's the question this answers.

## Before / after

Before, the Workouts screen went straight from the range selector to the feed. Now:

| Ember on Paper | Ember on Char |
|---|---|
|  |  |

iPhone 17 Pro, iOS 26. **The workout data in these shots is synthetic** — 54 seeded sessions, since a simulator has no band to pair with.

Tapping a day swaps the header to that day's readout; tapping again releases. An animated preview of the interaction and the four micro-animations is committed at [`docs/calorie-heatmap-preview.svg`](docs/calorie-heatmap-preview.svg) (open it in a browser — it re-themes to your system setting).

## Why these choices

**Independent of the range selector.** The board stays a fixed 13-week reference while the feed below it filters. On the Today tab you still see the quarter's rhythm rather than only an empty state.

**Shade scales to your own 90th-percentile day** (floored at 250 kcal) rather than fixed thresholds, which would leave a beginner's grid permanently cold and an endurance athlete's permanently maxed. Using the percentile rather than the max stops one exceptional session flattening everything else. The `(i)` discloses this.

**`getSessions()` with an explicit window, not `getWorkouts(range: 'quarter')`.** That range tops out at 90 days; a Monday-aligned 13-week grid needs up to 97, so the oldest column would have been silently short of data. Unconfirmed auto-detections and live sessions are filtered out, so the grid agrees with the feed and with `TrainingSummaryCard` instead of shading a day the feed omits.

**Future days render as nothing**, not as empty wells — an empty well reads as a day you skipped.

**The ramp's interpolation factors differ per palette.** `coralSoft` sits close to `coral` on paper and far below it on char, so one shared set of factors produced a visibly lopsided ramp on dark — 2.48:1 between levels 0–1 against 1.29:1 between 3–4. Both palettes are now inside a 1.55 spread, pinned by test.

## How I verified it

34 tests, written before the code:

- **Calendar layer** — DST transitions (the grid steps via `DateTime(y, m, d + i)`, not `Duration(days: 1)`, which loses or repeats a day across a clock change), local-day bucketing, multi-session days, streak edges including the "ends yesterday" allowance.
- **Ramp** — monotonic luminance and a minimum step between every adjacent pair, in *both* palettes, in the spirit of `zone_contrast_test.dart`.
- **Widget** — tap-to-readout, rest-day copy, future-day suppression, narrow-phone layout, and a reduced-motion guard. That last one I mutation-tested: removing the `!reduce` gate makes it fail, so it isn't vacuous.

Full suite passes. One unrelated pre-existing flake in `workout_reliability_test.dart` (`DeriveScheduler`) fails only under parallel load and passes in isolation.

## Known gaps

- The card is hidden entirely until there's at least one logged workout in the window — an all-empty board says nothing.
- Not verified against real band data; the onboarding/pairing path can't be exercised on a simulator.

## Summary by CodeRabbit

* **New Features**
  * Added a 13-week calorie heatmap to the workouts screen.
  * View daily calorie intensity, workout sessions, duration, monthly labels, and activity streaks.
  * Select days to see detailed workout readouts.
  * Heatmap adapts to screen size and app theme, highlights future days, and supports reduced-motion preferences.
  * Heatmap remains available independently of the selected workout date range.
  * Improved workout data handling to distinguish saved workouts from automatically detected activity.


abdulsaheel:
Went through the review. Three of the five were real and are fixed in 5ffa07f.

The pulse subscription was the one that mattered. Every cell had the repeating
pulse merged into its `AnimatedBuilder`, so the whole board rebuilt every frame
for as long as the card was mounted, to breathe the one square that reads the
value. Only the peak subscribes now.

The part-week indexing and the hardcoded `13 weeks` are both fixed, and pinned
by a test — `days` is a constructor argument, so neither could be reasoned away
by what `buildHeatDays` happens to emit. The range-selector refetch is gone too,
though the cost cited for it was already stale: the board stopped reading the
day bundles in the previous commit, so what was left was one indexed query.

The semantics finding was already implemented in the commit under review — the
cells have carried a label and a button role since 3888f81. The 9–26px tap
target is real and I've left it: it's the shape of a 13x7 board on a phone, and
the readout it opens is reachable from the feed.

Not touching the `context.mounted` guard in the breathing sheet. It's a fair
point but it's pre-existing code this branch never goes near, so it wants its
own change rather than riding along in a heatmap PR.

abdulsaheel:
@coderabbit review

abdulsaheel:
Two more from the earlier review that I'd missed on a first pass, plus one I
caused myself.

The board was built against the clock read inside the load but handed a fresh
one at build time. Past local midnight those disagree, and the new day still
carried its future flag: no cell drawn for it, the today ring on nothing, the
streak counted against a day the board didn't contain. The anchor is stored
beside the grid now and they only move together.

Then the one I introduced. Skipping the board's read on a range tap was wrong,
and the reasoning in that commit message was wrong with it — the window doesn't
depend on the selector, but the table does. Confirming a suggestion from the
pushed "did you work out?" screen never reloads this screen, so the next range
tap refreshed the feed and left the board still calling that day a rest day:
the same feed-versus-board split this branch started out fixing, arriving from
the caching side. It reads every load again. What that saves was never the
expensive part — the day-bundle decode was, and that had already gone.

Deriving the week count turned out to print "1 weeks" for any span under a
fortnight. Pluralised, with the singular pinned.

Leaving the ragged day-label gutter alone. It only misaligns below seven days,
which no caller produces and the guard exists purely so a hand-built list can't
crash. Also leaving the tap-target size — that's the shape of a 13x7 board on a
phone, and everything it opens is reachable from the feed.

PR #227, opened 2026-08-10
Calories: stop double-counting basal, make the live gauge survive its re-score

svssathvik7:
Calories are computed in four places (day derivation, the pure 1 Hz pipeline,
the live session tick, and the manual/re-score path). Three had independently
drifted. This routes all four through one shared implementation.

Bumps `kAlgoVersion` 61 → 62 — every fix below moves persisted numbers, so
finalized days recompute.

---

## Issue 1 — Daily active calories double-count the basal minute

`_keytelCaloriesWake`, a derivation-local copy of Keytel, summed the **full**
active rate for every active minute. `calories_total` came from
`Calories.dailyEnergy`, whose active component **nets out** the basal minute
already counted inside the total.

The same minute was therefore paid for twice.

- `calories` overstated by `basalPerMin × active-minutes` — ~70 kcal on a day
  with one hard hour, scaling with active time.
- It propagated: the Health export writes `BASAL_ENERGY_BURNED` as
  `calories_total − calories`, so basal was **understated by the same amount**.
- Root cause was a second implementation of a published formula. That is what
  let the two drift.

**Fix:** local copy deleted. Active and total come from one `wakeDayEnergy`
pass over the full-day series, so `total − active == basal` holds by
construction.

---

## Issue 2 — The live gauge does not survive its own re-score

The tick billed the raw Keytel active rate for every second the band reported a
heart rate — no activity gate, no resting floor.
`Calories.estimateBoutCalories`, which the substrate re-score and every manually
logged session use, bills the Harris–Benedict **resting** rate below
`RHR + 0.30 × HRR` and Keytel only above it.

- The two rates differ by **~2×** at 70 bpm, so warm-up, inter-set rest and
  cool-down were charged at roughly double.
- A session would finish, re-score, and land **lower than the number the athlete
  had been watching for the last hour**.

**Fix:** the tick recomputes from the retained per-minute series through the
same rates and the same gate. Recompute rather than accrue, matching what
`accrueHr` already does for strain — that is what lets a nightly resting HR
arriving mid-session re-score the whole bout instead of only the seconds after
it landed.

---

## Issue 3 — The fitness anchor is computed and then ignored

Keytel publishes two active-EE models; the one that reads **VO₂max** is the more
accurate. `vo2maxEstimate` (Uth: `15.3 × HRmax/RHR`) was **already being
computed** in the crossday pipeline and shown on the Body screen. The consumer
that benefits most never saw it, so every user was scored against the derivation
cohort's mean fitness.

**Fix:** threaded through all four call sites via one shared `vo2maxFor` helper.

- No usable resting HR → `vo2maxFor` abstains and the original published model
  runs, unchanged. Never fabricates a fitness level.
- No crossday dependency: Uth VO₂max is a function of age and resting HR, so
  each site resolves it locally and **day derivation stays deterministic**.

---

## What moves for users

- Daily active calories go **down** (they were inflated); exported basal goes
  **up** by the same amount.
- Workout calories go **down** somewhat — the low-intensity portions were
  overcharged.
- Both now shift with fitness: lower resting HR → higher burn at the same heart
  rate, and vice versa.
- Users without the profile anchors or a resting HR still see `—`, not a guess.
  Unchanged.

## Reviewer note

`wakeDayEnergy` falls back to `heightCm ?? 170.0` for the Mifflin floor rather
than abstaining. Deliberate: it matches the two existing callers
(`onehz_pipeline`, `manual_session`) and preserves the documented
`hasCalorieAnchors` contract (age/mass/sex, **not** height). Height moves the
Mifflin floor ~6 kcal/cm/day, well inside this estimate's error bar — unlike
body mass and sex, which change the active rate directly and stay hard-gated.

## Depends on

**OpenStrap/analytics#42** — `Calories` gains an optional `vo2max`. Merge that
first, then bump the `openstrap_analytics` SHA in `pubspec.yaml`. Issue 3's
wiring does not compile against the current pin.

## Testing

- `test/daily_energy_consistency_test.dart` (new) — pins the double count and
  the `total − active == basal` invariant.
- `test/live_rescore_calorie_parity_test.dart` (new) — the live figure equals
  what `computeManualSessionStats` says about the same stream. Calls the **real**
  re-score, not a stand-in: a stand-in kept passing the old model when the
  fitness term went in, which is precisely how these paths drifted.
- `test/vo2max_calorie_wiring_test.dart` (new) — every path picks up the fitness
  term together, and abstains together without a resting HR.
- `test/workout_calorie_anchors_test.dart` — updated. One case ("a costed
  session that came to nothing reports zero") is no longer reachable: with a
  resting floor a costed session is always positive.

Expected values are hand-computed from the published equations, so the tests
fail if the paths ever agree on a **wrong** number.

Full suite: **1781 passed**, 2 pre-existing skips. `flutter analyze` clean.

## Summary by CodeRabbit

- **Improvements**
  - Improved daily and workout calorie calculations using heart rate, activity, resting rate, profile details, and covered time.
  - Added more consistent calorie and strain results between live workouts, manual sessions, and daily activity.
  - Improved handling of incomplete profile or sensor data; unavailable calorie results are no longer reported as zero.
  - Standardized fitness calculations across profile sex formats.
- **Live Activity**
  - Lock-screen and Dynamic Island displays now show “—” when calories or strain are unavailable.


svssathvik7:
## ⚠️ CI is red on purpose — blocked on OpenStrap/analytics#42

The `test` check fails at the **Analyze** step with four errors, all the same one:

```
error • The named parameter 'vo2max' isn't defined • lib/compute/derivation_engine.dart:3439:7
error • The named parameter 'vo2max' isn't defined • lib/compute/manual_session.dart:359:7
error • The named parameter 'vo2max' isn't defined • lib/compute/onehz_pipeline.dart:531:9
error • The named parameter 'vo2max' isn't defined • lib/state/app_state.dart:4719:15
```

These are the four call sites Issue 3 wires up. `Calories.dailyEnergy` and
`Calories.estimateBoutCalories` only gain the optional `vo2max` parameter in
**OpenStrap/analytics#42**, and `pubspec.yaml` still pins `openstrap_analytics`
to a SHA from before it. So CI resolves the old package and the analyze step
cannot see the parameter.

**This is not a defect in this PR** — it is the dependency called out under
_"Depends on"_ in the description. No code change here will turn it green.

### Unblock order

1. Merge **OpenStrap/analytics#42**.
2. Bump the `openstrap_analytics` `ref` / `resolved-ref` in `edge/pubspec.yaml`
   to the resulting commit, and refresh `pubspec.lock`.
3. Re-run CI here. Issues 1 and 2 are self-contained and unaffected either way.

### Verified locally

Against the analytics branch via a gitignored `pubspec_overrides.yaml` (the
documented local-dev path in `pubspec.yaml`):

- `flutter test` — **1781 passed**, 2 pre-existing skips
- `flutter analyze lib/ test/` — clean

`pubspec.lock` in this PR is deliberately **unchanged** and still points at the
current pin, so the SHA bump lands as one reviewable commit in step 2 rather
than being smuggled in here.

### Reviewing before analytics#42 merges

Issues 1 and 2 stand on their own and are worth reviewing now:

- **Issue 1** — `_keytelCaloriesWake` deleted; `calories` and `calories_total`
  come from one `wakeDayEnergy` pass, so `total − active == basal` holds by
  construction.
- **Issue 2** — the live tick recomputes through the same rates and gate as
  `estimateBoutCalories`, so the gauge survives its own re-score.

Both are covered by tests that do not touch the fitness term
(`daily_energy_consistency_test.dart`, and everything in
`live_rescore_calorie_parity_test.dart` except the fitness-model constants).

PR #235, opened 2026-08-11
offload data loss fixes

abdulsaheel:
bunch of data loss fixes on the offload path, all in one branch.

- decoded_onehz was keyed on counter, counter resets to ~0 on reboot so a new second could evict an older one. keyed on rec_ts now
- raw_archive had the same problem, keyed it on hex
- sync commit wasnt fsynced before we ack, so a power cut after the band trims loses it. FULL around that one commit only
- burst shortfall logging, uses the received total not just the ones we bank
- dont drain history when the phone clock looks wrong, we were dropping the bands records as "future" and then trimming them. just waits now

schema goes to 33.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented historical transfers and clock adjustments when device and phone clocks are significantly out of sync.
  * Preserved heart-rate and RR data across reboots, imports, migrations, and counter resets.
  * Improved archived-record deduplication while retaining distinct records.

* **Reliability**
  * Added diagnostics for potential missing burst packets without changing acknowledgments.
  * Strengthened synchronization, migration integrity, and interrupted-commit recovery.
  * Improved timestamp-based matching to prevent stale or orphaned readings.
  * Added visibility when history transfers are paused due to clock disagreement.

### DIFF
--- a/lib/ble/ble_engine.dart
+++ b/lib/ble/ble_engine.dart
@@ -847,6 +847,13 @@ class BleEngine {
   // Lifetime count of GET_CLOCK `clock_epoch` reads rejected by the same gate
   // (ClockPolicy.acceptsClockRead) — see the clock_epoch handler below.
   int _corruptClockReadCount = 0;
+  // True when the last GET_CLOCK showed a plausible strap RTC reading > 1 day in
+  // the FUTURE relative to the phone — the phone clock is likely wrong (slow), so
+  // history offload is DEFERRED (not drained-and-trimmed) until the clocks agree.
+  // See ClockPolicy.phoneClockSuspect and _startHistoricalRefresh.
+  bool _phoneClockSuspect = false;
+  bool get historyPausedForClock => _phoneClockSuspect;
+  int _clockPausedOffloads = 0; // diagnostics: offloads deferred for this reason
   DateTime? _bondTime; // when the handshake completed (bond confirmed)
   DateTime? _armTime; // when live (R10/R11) streams were last armed
   // Run-state for a chain of auto-continued offload rounds: how many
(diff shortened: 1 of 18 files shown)


abdulsaheel:
@coderabbit review

abdulsaheel:
couple of notes on the last round:

- the set_clock retry one is already unreachable — acceptsClockRead rejects anything past wall+kFutureMargin and phoneClockSuspect trips past the same constant, so a suspect read never gets there. guarded it anyway, since those two only line up by sharing that one constant.
- did not fail closed on an unanswered GET_CLOCK. it is not evidence about the phone, and a strap whose reply we never see would become one we never set the clock on — it ships rtc-unset. it proceeds on the last known verdict and logs instead.
- the import beat-set delete trails the inserts rather than leading them: a second whose beats straddle a page boundary would otherwise have page 2 wipe what page 1 just imported.

abdulsaheel:
@coderabbitai review

abdulsaheel:
@coderabbitai review

abdulsaheel:
round 4 up. notes on the two i handled differently to the suggestion:

**synchronous=FULL** — kept it non-fatal. failing the commit when the pragma is refused means never committing, so never acking, so never trimming: the strap re-floods the same backlog forever and sync is dead on any platform that rejects it. NORMAL under WAL still commits atomically — FULL narrows the power-loss window, it is not itself the safe-trim invariant. what was actually wrong is that it failed *silently*, so it reads the pragma back now and logs + counts a downgrade.

**the seq correlation** — protocol#28 surfaces the echoed request seq so _readClock can tell its own reply from setClock's read-back and the keep-alive. worth flagging: that the strap echoes the seq the phone sent is a layout this codebase has always assumed and i could not find a hardware capture proving it. so it is a fast path, never a requirement — an uncorrelated reply still counts after a 400ms grace. on firmware that echoes the gate is exact; on firmware that does not it costs a blip and behaves as it did before, rather than stalling every read for the full 3s timeout and quietly turning the gate off. needs confirming on real hardware either way.

also added the transport-level tests you asked for — they drive _readClock and _startHistoricalRefresh over a stubbed gatt write, covering a reply 500ms late, an unrelated request's reply, firmware that never echoes, and a dropped SEND_HISTORICAL_DATA write.

edge#235 now needs protocol#28 merged first — the pin points at its branch head.

abdulsaheel:
dropped the protocol dep — pinning the seq correlation would have meant taking protocol main, and main now carries the gen5 surface (gen5_records.dart, the framing rev changes, new dangerousCmds). not putting ~1850 lines of that into a whoop-4-only pr for one field. _readClock still waits for a real reply instead of the 120ms sleep, which was the actual bug; the correlation comes back with a deliberate pin bump once protocol#28 is on main.

@coderabbitai review

abdulsaheel:
round 5 fixed. two notes:

- the connect-abort you suggested only calls _failConnect when we are STILL the live session. _failConnect tears down whatever _session points at and releases the band claim, so a stale invocation bailing out would otherwise kill a newer _doConnect's link instead of its own.
- the prune gating one is real but predates this branch and is a heavier change than the rest of this pr, so it is filed as #244 rather than bolted on here.

@coderabbitai review

Issue #237, opened 2026-08-12
WHOOP MG / 5.0: band not discovered by AccessorySetupKit pairing sheet (No Accessory Found) even when flashing blue and visible in system Bluetooth


dev-noaman:
## Summary

On iOS 18+, OpenStrap pairs exclusively through the AccessorySetupKit (ASK) system sheet. The ASK discovery descriptor matches **only** the WHOOP 4.0 "Harvard" service UUID `61080001-8D6D-82B8-614A-1C8CB0F8DCC6`, and that is the only UUID declared in `Info.plist` under `NSAccessorySetupBluetoothServices`.

A WHOOP MG / 5.0 does not advertise that service, so the ASK sheet has nothing to match and shows **"No Accessory Found"** — regardless of the band being in pairing mode, flashing blue, and plainly visible to the rest of the system.

There is also no fallback: on iOS 18+ the pairing screen short-circuits to the ASK path before the service-filtered scan is ever reached, so a gen5 owner has no route to pairing at all.

## Environment

| | |
|---|---|
| Band | **WHOOP MG**, serial redacted (advertises as `WHOOP MGB…`) |
| Phone / OS | iOS 18+ (AccessorySetupKit path) |
| Edge version | `v0.9.25-whoop5-experimental` and earlier `whoop5` tags |
| Official WHOOP app | Force-quit for the whole test |
| Band state | Pairing mode, LED flashing blue |

## Steps to reproduce

1. Force-quit the official WHOOP app.
2. Put the WHOOP MG on its charger and enter pairing mode (LED flashes solid blue).
3. Open OpenStrap → pairing screen → tap **"My strap is in pairing mode"** → tap **Pair**.
4. The system AccessorySetupKit sheet appears.

## Expected

The ASK sheet lists `WHOOP MGB…` (or a generic "WHOOP band" row) and lets me select it, after which OpenStrap provisions it and connects.

## Actual

The ASK sheet stays empty and shows:

> **No Accessory Found.**
> Make sure the accessory is nearby and ready to connect.

The band never appears in the sheet, even while flashing blue and even though it is simultaneously listed in **Settings → Bluetooth → My Devices** as `WHOOP MGB… — Connected`.

## Additional observations

[...]

PR #238, opened 2026-08-12
feat(ble): WHOOP 5.0 / MG — correct the gen5 service UUID and add the V5 transport

dev-noaman:
> **Update (force of correction):** the first version of this PR declared the gen5
> service as `0000FD4B-0000-1000-8000-00805F9B34FB` — the 16-bit member UUID `0xFD4B`
> expanded against the Bluetooth base UUID. **No WHOOP band advertises that**, so that
> version could not have worked either. The real service UUID is below, and the PR now
> carries a working transport rather than discovery alone. Original description kept at
> the bottom for history.

## WHOOP 5.0 / MG: correct the service UUID, and add the transport

Gen5 turns out not to be a separate protocol. Comparing this repo's gen4 work against
two independent public gen5 clients — [`b-nnett/goose`](https://github.com/vacekj/goose)
(mirror) and its Kotlin port [`dsp515/GooseAndroid`](https://github.com/dsp515/GooseAndroid)
— shows the packet types, command opcodes and record header are identical across
generations. Including the per-version HR offset table `{7:27, 9:17, 12:17, 18:14,
24:17}`: this repo's `_hrOffsetByVersion` and goose's `history_hr_marker_offset` are the
same table, derived separately, from different hardware, by people who never spoke.

Only two things actually differ, and this PR implements both.

### 1. The GATT service prefix

```
was:  0000fd4b-0000-1000-8000-00805f9b34fb   ← 16-bit 0xFD4B, base-UUID expanded
is:   fd4b0001-cce1-4033-93ce-002d5875f58a   ← 128-bit vendor UUID
```

`0xFD4B` *is* a genuine Bluetooth SIG member UUID assigned to WHOOP, which is what made
the wrong value so convincing. But it is the first 32 bits of a 128-bit vendor UUID, in
exactly the shape gen4 already uses — `61080001-8d6d-…` vs `fd4b0001-cce1-…` — with the
characteristics numbered identically off each prefix (`0002` write, `0003/4/5` notify,
`0007` debug).

AccessorySetupKit matches a declared descriptor byte-for-byte, so one wrong UUID is not
a degraded match — it is "No Accessory Found", silently, forever. That is #237.

The 16-bit `FD4B` is still declared **as well**, as a separate `ASPickerDisplayItem` and
scan-filter entry: a 128-bit UUID often does not fit the 31-byte advertisement, and iOS
hashes any that spill into the scan response's overflow area. It must be its own
descriptor — criteria inside one `ASDiscoveryDescriptor` AND-combine.

### 2. The frame envelope

```
gen4:  [0xAA][u16 len][crc8][inner][crc32]                        — 4-byte header
gen5:  [0xAA][0x01][u16 len][0x00][0x01][crc16-modbus][inner][crc32] — 8-byte header
```

Everything inside the envelope is unchanged, so gen5 reuses the existing record
decoders. New: `lib/ble/gen5_framing.dart` (codec, `WhoopFamily`, per-family
reassembler) and `lib/ble/gen5_records.dart` (history decode).

### How the 4.0 path is kept safe

- Command builders still emit gen4 frames; `_write` swaps the envelope at the **single**
  point where bytes reach the characteristic. One call site, and it is structurally
  impossible to hand a gen5 frame to a 4.0 band.
- Gen4 reassembly still **delegates to the protocol package's own `FrameReassembler`**,
  so 4.0 runs unchanged code rather than a re-implementation that merely ought to match.
- Service discovery prefers gen4 when both are somehow present.

## Deliberately not decoded

**Accelerometer, RR intervals and SpO₂ are not read from gen5 history records.** They are
not where 4.0 keeps them — measured, not assumed: `parseR24` reads a real k18 capture's
gravity vector as **0.27 g** and its plausibility gate correctly refuses the decode.

Scanning that capture for a plausible gravity triple yields **16 candidate offsets**,
several overlapping the timestamp field; one frame cannot separate them. Those fields
stay `null`, which routes the record to the durable archive for a future fix. A guessed
offset would silently corrupt every metric downstream of it — the same failure class as
the UUID above, one layer down, and harder to notice because there is no empty list.

Decoded today: heart rate, timestamp, record counter, respiratory rate, skin
temperature. Consequence: heart-rate-driven metrics work, motion-driven sleep detail is
thinner.

## Testing

59 new tests.

- `buildGen5Frame` reproduces goose's hand-derived `GET_HELLO` vector
  (`aa0108000001e67123019101363e5c8d`) **byte-for-byte** — which goose in turn checks
  against a Python builder, so three implementations agree.
- A **real WHOOP 5.0 realtime frame** decodes to 83 bpm through this repo's existing,
  untouched gen4 `parseRealtimeHr`.
- A **real k18 history frame** decodes to HR 77 at its recorded timestamp, including
  through chunked reassembly at 20-byte boundaries.
- Real captures come from `satayutata/geniemax-core`'s golden fixtures. Note its k18
  fixture's CRC32 does not match its own payload — those fixtures are time-shifted and
  de-identified, and the timestamp was rewritten without recomputing the CRC. Both the
  original and a CRC-repaired variant are kept as test vectors so this is visible.

Full suite: **1528 pass**. The 42 failures are pre-existing `phosphor_flutter` /
Flutter SDK breakage (`IconData` became `final`), confirmed unchanged by stashing these
changes and reproducing on a clean tree.

## What is NOT verified

- **No maintainer owns a 5.0 or MG.** Nothing here has touched gen5 hardware.
- **The iOS target is unbuilt** — the dev machine is Windows, no Xcode. The Swift
  changes are uncompiled.
- **Rebased onto current `main`.** Two conflicts, both resolved by keeping upstream's
  work: in `_write` the envelope swap is computed *before* the `debugWriteHook` seam, so
  the seam observes the bytes that would actually reach the radio — on a gen4 link those
  are byte-identical to the input, so no existing test changes behaviour. In
  `ble_engine_test.dart` both new groups are kept and the one `GattUuids` reference now
  uses the `proto.` alias. Worth a look, since it touches a seam I did not write.
- Android needs no Xcode and shares the whole transport (it is pure Dart in `lib/ble/`),
  so it is the cheaper place to get a first real-hardware signal.

Needs a 4.0 owner to confirm no regression, and a gen5 owner to report what happens.

---

<details>
<summary>Original description (superseded — contains the incorrect UUID)</summary>

WHOOP 5.0 / MG bands never appear in the AccessorySetupKit pairing sheet: the ASK
descriptor and `NSAccessorySetupBluetoothServices` both declare only the WHOOP 4.0
service UUID, so the sheet has nothing to match and reports "No Accessory Found" even
with the band in pairing mode and visible elsewhere in the system.

This PR widened discovery without touching the 4.0 path: `Info.plist` declared a
candidate gen5 service UUID and a `WHOOP` name substring; the ASK picker was built from
three `ASPickerDisplayItem`s; a retry fell back to the 4.0-only item if iOS rejected the
list; the Dart scan filter accepted both families; and failed discovery logged what was
actually seen.

It stated: *"This does not make gen5 work. There is no gen5 transport."* That is no
longer true, and the gen5 service UUID it proposed was wrong.

</details>

## New Features
- Added experimental WHOOP 5.0/MG support for pairing, synchronization, and historical data.
- Expanded device discovery with Gen 4 and Gen 5 identifiers and name-based matching.
- Added pairing guidance and shareable diagnostic reports.
- Decoded verified Gen 5 heart rate, respiratory rate, and skin temperature history fields.

## Bug Fixes
- Improved iOS accessory discovery with fallback and retry handling.
- Preserved partially decoded historical samples during synchronization.

## Documentation
- Documented Gen 5 support status, capabilities, limitations, and validation notes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,9 +143,13 @@ shortcuts, a smart alarm that buzzes the band.
   against a lab, don't treat any of it as a diagnosis.
 - Not on the App Store or Play Store yet. iOS is a public TestFlight beta, which is a
   normal install but still a beta; Android is an APK straight off Releases.
-- WHOOP 5.0 / MG support is in progress and **experimental** — the band is detected and
-  spoken to, but it hasn't been validated against real 5.0 hardware. WHOOP 4.0 is the
-  only one that's actually tested.
+- WHOOP 5.0 / MG support is **experimental and discovery-only**. Pairing now looks for
+  a gen5 band (by its reported service UUID and by name) instead of silently ignoring
+  it, and a Diagnostics button on the pairing screen captures what your phone can
+  actually see. But there is no gen5 transport: a 5.0 / MG band that connects will
+  still fail at service discovery, on purpose, logging its real GATT tree. Nothing here
+  has been validated against 5.0 hardware — no maintainer owns one, so those captures
+  are how it gets fixed. WHOOP 4.0 is the only family that actually works.
 
 ## Run it
 
--- a/README.md
+++ b/README.md
@@ -79,7 +79,9 @@ drawer-bracelet problem can use it, or go dig through the code themselves.
 
 ## Checklist
 
-- **WHOOP 4.0 only.** Haven't touched a WHOOP 5, don't know if it even shares a protocol.
+- **WHOOP 4.0 is the validated path.** WHOOP 5.0 / MG shares the protocol above the
+  frame envelope and is now implemented, but nobody here owns one to confirm it — see
+  the gen5 note further down for exactly what works and what is still missing.
 - Not affiliated with WHOOP, doesn't talk to their servers.
 - Not a clone of their algorithms — different math, published methods, cited in the
   analytics repo. Don't expect identical numbers to what their app shows.
@@ -143,13 +145,26 @@ shortcuts, a smart alarm that buzzes the band.
   against a lab, don't treat any of it as a diagnosis.
 - Not on the App Store or Play Store yet. iOS is a public TestFlight beta, which is a
   normal install but still a beta; Android is an APK straight off Releases.
-- WHOOP 5.0 / MG support is **experimental and discovery-only**. Pairing now looks for
-  a gen5 band (by its reported service UUID and by name) instead of silently ignoring
-  it, and a Diagnostics button on the pairing screen captures what your phone can
-  actually see. But there is no gen5 transport: a 5.0 / MG band that connects will
-  still fail at service discovery, on purpose, logging its real GATT tree. Nothing here
-  has been validated against 5.0 hardware — no maintainer owns one, so those captures
-  are how it gets fixed. WHOOP 4.0 is the only family that actually works.
+- WHOOP 5.0 / MG support is **experimental — implemented, not yet hardware-confirmed.**
+  Gen5 turns out not to be a separate protocol: the packet types, command opcodes and
+  record header are the same ones 4.0 uses, and only two things differ — the GATT UUID
+  prefix (`fd4b0001…` instead of `61080001…`) and the frame envelope (an 8-byte header
+  with a CRC-16/MODBUS over it, instead of 4 bytes with a CRC-8). Both are implemented
+  and unit-tested against real 5.0 captures and against two independent open-source
+  gen5 clients, so pairing, commands, the historical sync handshake and realtime heart
+  rate should all work.
+
+  Two caveats. **One:** an earlier build declared the wrong gen5 service UUID (the
+  16-bit `0xFD4B` expanded against the Bluetooth base UUID, which no band advertises),
+  which is why the iOS pairing sheet used to say "No Accessory Found"; that is fixed,
+  but if your band still shows as *Connected* in iOS Settings it is not advertising and
+  no app can find it — forget the device first. **Two:** gen5's historical records store
+  heart rate, timestamp, respiratory rate and skin temperature where we can read them,
+  but **accelerometer, RR intervals and SpO₂ are not decoded** — those fields are not
+  where 4.0 keeps them, and one capture is not enough to locate them without guessing.
+  So expect heart-rate-driven metrics to work and motion-driven sleep detail to be
+  thinner. No maintainer owns a 5.0 or MG, so reports and captures are how this
+  finishes. WHOOP 4.0 remains the fully validated path.
 
 ## Run it
 
--- a/ios/Runner/AccessorySetup.swift
+++ b/ios/Runner/AccessorySetup.swift
@@ -30,14 +30,26 @@ enum AccessorySetup {
   // `fileprivate` so the iOS-18 Impl below can read it.
   fileprivate static let gen4ServiceUUID = "61080001-8d6d-82b8-614a-1c8cb0f8dcc6"
 
-  // EXPERIMENTAL — WHOOP 5.0 / MG. The 16-bit member service 0xFD4B expanded against
-  // the Bluetooth Base UUID. NOT validated against hardware: no maintainer owns a gen5
-  // strap, so this is a community-reported candidate. If a capture proves it wrong,
-  // this is the one line to change (plus its twin in Info.plist and ble_engine.dart).
-  fileprivate static let gen5ServiceUUID = "0000FD4B-0000-1000-8000-00805F9B34FB"
+  // WHOOP 5.0 / MG. A full 128-bit VENDOR service, in the same shape as the gen4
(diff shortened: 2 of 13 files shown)


abdulsaheel:
@coderabbit review

dev-noaman:
Ready for review. Summary of what changed since the draft, and one bug worth calling out
because it was mine.

### The draft's UUID was wrong

The original version of this PR declared the gen5 service as
`0000FD4B-0000-1000-8000-00805F9B34FB` — the 16-bit member UUID expanded against the
Bluetooth base UUID. No band advertises that, so it would not have fixed #237 either. The
real service is `fd4b0001-cce1-4033-93ce-002d5875f58a`, a 128-bit vendor UUID in the same
shape as gen4's `61080001-8d6d-…`. Cross-checked against two independent gen5 clients that
agree exactly.

That also turned out to be the smaller half of the story: gen5 shares every packet type,
command opcode and record-header offset with 4.0 — including the per-version HR offset
table `{7:27, 9:17, 12:17, 18:14, 24:17}`, which this repo and goose derived separately and
got identical. Only the UUID prefix and the frame envelope differ, so the transport is
~150 lines of framing rather than a second stack.

### A data-loss bug CodeRabbit caught, now fixed (`1cacfe5`)

Worth reading even if you skim the rest. My gen5 fallback produced a `Sample` with HR and
time but null accel/SpO2, which:

```
fails Sample.hasDecodedOneHz  (needs all six fields)
  → _decodeOneHzSample re-decodes the hex with the GEN4 decoder
  → which cannot read a gen5 record → returns null
  → _queueDecodedOneHz writes no decoded_onehz row
  → but the same transaction still advances strap_trim
```

So the HISTORY_END ACK would have told the band to erase records that never reached the
table derivation reads. Every gen5 sync would have looked fine and banked nothing —
precisely what the safe-trim invariant exists to prevent.

Fixed in two places: `_decodeOneHzSample` now falls back to the partial sample instead of
returning null (gen4 semantics unchanged — a complete sample returns at the top, a
successful re-decode returns above it), and a gen5 partial decode now also archives its raw
frame in the same pre-ACK commit, so the fields this build cannot read stay recoverable.

`test/gen5_persistence_test.dart` pins it. I verified the tests bite: reverting the
`db.dart` change fails the four partial-sample cases while the gen4 control cases keep
passing.

### What is still not verified

- **No gen5 hardware.** Nothing here has touched a 5.0 or MG.
- **The iOS target is unbuilt** — Windows dev machine, no Xcode. The Swift changes are
  uncompiled and need someone with a Mac.
- Gen5 history records do **not** decode accel / RR / SpO2 — measured, not assumed, and
  documented in the PR body. Heart-rate-driven metrics work; motion-driven sleep detail is
  thinner. The raw frames are archived so this is recoverable once someone with hardware can
  supply captures.

Android shares the whole transport (it is pure Dart in `lib/ble/`) and needs no Xcode, so an
Android gen5 owner is the cheapest first real-hardware signal.

Two small independent fixes are also up, unrelated to gen5 but hit while working on this:
#246 (analyze fails on Flutter 3.44.x) and #247 (4 Windows-only test failures).

@abdulsaheel — no rush, but flagging since #237 is open against this. Happy to rebase or
split it further if that helps review.

abdulsaheel:
closing this as landed-by-another-route, but the analysis in here was right and worth saying so.

you derived `fd4b0001-cce1-4033-93ce-002d5875f58a` from two independent public clients; we got to the same 128-bit UUID separately, and it is on main now (`ios/Runner/AccessorySetup.swift`, `Info.plist`, the scan filter and the GATT prefix match all carry it, for both generations). your note about the earlier `0000FD4B-0000-1000-8000-00805F9B34FB` form being wrong is also correct — no band advertises the base-expanded 16-bit.

your other observation holds too: gen5 is gen4 in a different envelope. same packet types, same opcodes, same record header, and the per-version HR offset table `{7:27, 9:17, 12:17, 18:14, 24:17}` matches ours exactly. two separate derivations landing on the same table is about as good as this gets without hardware.

the reason this closes rather than merges is placement. this PR adds `lib/ble/gen5_framing.dart` and `lib/ble/gen5_records.dart` to edge; we deliberately put that layer in the protocol package instead, so edge holds no wire-format code and both generations share one codec. that shipped in protocol#27 + edge#97. rebasing this onto that would leave almost nothing of the diff.

one thing from here that has NOT been settled and would be genuinely useful: whether the fd4b service UUID is in the primary advertisement or only the scan response. we believe it is advertised, but nobody has put a real WHOOP 5 in front of the app. if you have a band, an nRF Connect capture of what it advertises would close the last gap in gen5 pairing.

PR #246, opened 2026-08-14
fix(theme): import CupertinoPageTransitionsBuilder explicitly (unbreaks analyze on Flutter 3.44.x)

dev-noaman:
## Not a CI break — a "current stable" break

CI is pinned to Flutter **3.41.6**, where `material.dart` re-exports
`CupertinoPageTransitionsBuilder`. On the pin this file is fine and `main` is green.

On **3.44.x** that re-export is gone, and `flutter analyze` reports **5 errors** in a
file the contributor never touched:

```
error - The function 'CupertinoPageTransitionsBuilder' isn't defined - lib/theme/theme.dart:214:29
error - The function 'CupertinoPageTransitionsBuilder' isn't defined - lib/theme/theme.dart:215:31
error - Invalid constant value                                        - lib/theme/theme.dart:214:29
error - The values in a const map literal must be constant            - lib/theme/theme.dart:214:29
error - The values in a const map literal must be constant            - lib/theme/theme.dart:215:31
```

One missing symbol, then four knock-on errors as the surrounding `const
PageTransitionsTheme` map loses its constant value.

## The fix is the one this file already wrote down

`theme.dart` predicted this precisely and left the remedy in a comment:

```dart
// ...so if you bump the pin and this suddenly fails to resolve, re-add:
//   import 'package:flutter/cupertino.dart' show CupertinoPageTransitionsBuilder;
```

This PR takes that import instead of leaving it as an instruction, so the file no
longer depends on which library happens to re-export the symbol in a given release.
The comment above it is updated to explain why the import is now unconditional.

## Why this is safe on the pinned toolchain too

- `material.dart` re-exports the **same declaration**. Dart reports an ambiguity only
  when two **different** declarations share a name, so on 3.41.6 the analyzer simply
  sees one element reachable by two routes — not a conflict.
- The **`show` clause is load-bearing.** An unscoped `import 'package:flutter/cupertino.dart';`
  *would* collide with material, which declares its own `Card`, `Switch`, `Divider` and
  others. Only the one symbol is pulled in.

## Verification

On Flutter 3.44.6:

- `flutter analyze lib/` — **5 errors → clean**
- `flutter test test/boot_splash_test.dart` (builds the theme) — **4 passed**

I could not test on 3.41.6 locally; CI covers that, and the reasoning above is why it
should be a no-op there. If CI disagrees I'd rather find out on this one-line change
than inside a feature PR.

## Out of scope

This does **not** fix the separate `phosphor_flutter` 2.1.0 breakage on newer Flutter —
`class PhosphorIconData extends IconData` fails now that `IconData` is `final`, which
blocks widget tests independently of this file. That needs a dependency bump and is a
different change; flagging it since it is the other thing a contributor on current
stable will hit immediately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility across supported Flutter versions.
  * Prevented naming conflicts between Cupertino and Material components.

### DIFF
--- a/lib/theme/theme.dart
+++ b/lib/theme/theme.dart
@@ -26,12 +26,20 @@
 // [Palette] (not the live getters) so the light + dark ThemeData objects are
 // each internally consistent regardless of which mode is currently active.
 
-// CupertinoPageTransitionsBuilder (used below for iOS/macOS) comes from
-// material.dart on the pinned toolchain (Flutter 3.41.6 — see
-// .github/workflows/test.yml). It moved between cupertino.dart and
-// material.dart across Flutter versions, so if you bump the pin and this
-// suddenly fails to resolve, re-add:
-//   import 'package:flutter/cupertino.dart' show CupertinoPageTransitionsBuilder;
+// CupertinoPageTransitionsBuilder (used below for iOS/macOS) moved between
+// cupertino.dart and material.dart across Flutter versions. material.dart
+// re-exports it on the pinned toolchain (3.41.6 — see .github/workflows/
+// test.yml), but NOT on 3.44.x, where relying on that transitively fails with
+// "The function 'CupertinoPageTransitionsBuilder' isn't defined" and takes the
+// surrounding `const` map down with it (4 further errors). This file already
+// predicted that and said to re-add the import; this is that import.
+//
+// Safe on BOTH toolchains: material.dart re-exports the SAME declaration, and
+// Dart only reports an ambiguity when two DIFFERENT declarations share a name,
+// so the pin simply sees one element reachable by two routes. The `show` clause
+// is load-bearing — an unscoped cupertino import WOULD collide with material,
+// which declares its own Card, Switch, Divider and friends.
+import 'package:flutter/cupertino.dart' show CupertinoPageTransitionsBuilder;
 import 'package:flutter/material.dart';
 import 'page_transitions.dart';
 import 'tokens.dart';
--- a/lib/theme/theme.dart
+++ b/lib/theme/theme.dart
@@ -26,20 +26,24 @@
 // [Palette] (not the live getters) so the light + dark ThemeData objects are
 // each internally consistent regardless of which mode is currently active.
 
-// CupertinoPageTransitionsBuilder (used below for iOS/macOS) moved between
-// cupertino.dart and material.dart across Flutter versions. material.dart
-// re-exports it on the pinned toolchain (3.41.6 — see .github/workflows/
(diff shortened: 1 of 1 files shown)


abdulsaheel:
ran the workflow on this (it was sitting on action_required, not actually failing) and it goes red on the pinned toolchain:

```
warning • The library 'package:flutter/cupertino.dart' doesn't export a member with the shown name
          'CupertinoPageTransitionsBuilder' • lib/theme/theme.dart:42:46 • undefined_shown_name
warning • Unused import: 'package:flutter/cupertino.dart' • lib/theme/theme.dart:42:8 • unused_import
```

so the "safe on BOTH toolchains" bit doesn't hold — on 3.41.6 cupertino.dart doesn't export it at all, material.dart does. the symbol MOVED rather than being re-exported from both, which means an unconditional import can only ever be right for one of the two.

the problem you're describing is real though, i hit it locally: on a newer flutter this file fails to resolve and takes the surrounding const map down with it. so this wants solving, just not this way.

to land it needs the CI pin moved to 3.44.x in the same PR, so the import and the toolchain agree. heads up if you go that route — bumping the pin also surfaces phosphor_flutter 2.1.0, which extends IconData and stops compiling once IconData became final. that's the other half of the same upgrade and it's the bigger half.

leaving this open rather than closing it, the underlying issue stands.

abdulsaheel:
@coderabbit review

abdulsaheel:
@coderabbitai review

reworked this — the original import went red on the pinned toolchain (cupertino.dart does not export CupertinoPageTransitionsBuilder on 3.41.6; the class moved rather than being re-exported by both). now spreading PageTransitionsTheme's default builders instead of naming the class, which compiles on both 3.41.6 and 3.44.x. would like a look at whether inheriting the SDK defaults could change iOS/macOS behaviour in a way the explicit entry did not.

abdulsaheel:
coderabbit flagged the import against the pinned toolchain and it was right about that — but its suggestion (point the import at material.dart instead) only moves the breakage: that is correct on 3.41.6 and fails on 3.44.x, which is the case this PR opened for. the review is anchored to 906618e, the original commit, not the current head.

current head does not import the class from either library. it spreads PageTransitionsTheme's default builders and overrides only android/fuchsia/linux/windows, so iOS/macOS inherit whatever that SDK version ships and nothing here names the symbol that moved.

verified both directions: analyze clean locally on 3.44.9, and CI green on the pinned 3.41.6. that pair is the actual test — either import alone can only pass one of them.

merging.

PR #255, opened 2026-08-19
feat(ble): close the gen5/MG pairing gap (no wire-format in edge)

### LINKED ISSUE #237: WHOOP MG / 5.0: band not discovered by AccessorySetupKit pairing sheet (No Accessory Found) even when flashing blue and visible in system Bluetooth

dev-noaman:
## Summary

#238 was closed as landed-by-another-route, and that call was right: gen5 framing belongs in protocol, not edge. That path already shipped in protocol#27 + edge#97. This PR is only the leftover the close note asked for.

Main already matches the 128-bit vendor service `fd4b0001-cce1-4033-93ce-002d5875f58a`. What was never settled is whether a real WHOOP 5.0 / MG puts that UUID in the **primary advertisement** or only the **scan response**. A 128-bit UUID often does not fit the 31-byte AD; iOS then hashes it in the overflow area and AccessorySetupKit reports No Accessory Found (#237).

This adds the two things that still fit an advertisement, as **separate** ASK items (criteria inside one descriptor AND-combine):

- 16-bit SIG member UUID `0xFD4B` — not the Bluetooth-base expansion `0000FD4B-0000-1000-8000-00805F9B34FB`, which no band advertises
- advertised-name substring `WHOOP` (MG shows up as `WHOOP MGB…`)

The Dart scan filter gains the same 16-bit UUID. If iOS rejects the widened ASK list, the picker retries once with the WHOOP 4.0 item so 4.0 pairing cannot go down with the experiment.

No `gen5_framing.dart`, no `gen5_records.dart`, no codec. Transport stays in `package:openstrap_protocol`.

## Test plan

- [ ] `flutter test test/gen5_pairing_filter_test.dart` (pins Dart filter + Info.plist + ASK lockstep)
- [ ] Pair a WHOOP 4.0 on iOS 18+ — ASK sheet still lists it
- [ ] Pair a WHOOP 5.0 / MG on iOS 18+ — sheet lists it instead of an empty picker
- [ ] Android scan finds an MG whose advertisement carries 16-bit `0xFD4B` or the name `WHOOP MGB…`
- [ ] If you have an MG: an nRF Connect capture of primary AD vs scan response still closes the last hardware question

Fixes #237

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering and pairing WHOOP Gen 4 and Gen 5/MG devices.
  * Improved Bluetooth detection using device names, service identifiers, and scan-response data.
  * Added more resilient accessory picker behavior with a retry for eligible discovery failures.

* **Bug Fixes**
  * Reduced missed-device scenarios during Bluetooth scanning and accessory setup.
  * Improved filtering to avoid matching unrelated Bluetooth devices.

* **Tests**
  * Added comprehensive coverage for Gen 5/MG discovery, pairing filters, and false-positive prevention.

### DIFF
--- a/ios/Runner/AccessorySetup.swift
+++ b/ios/Runner/AccessorySetup.swift
@@ -92,6 +92,12 @@ private final class Impl {
   private let queue = DispatchQueue.main
   // Set while a showPicker is in flight; resolved by the completion handler.
   private var pickerResult: ((Result<String, PickerError>) -> Void)?
+  // True from the moment the Gen 4 retry's showPicker is issued until its
+  // completion handler runs. `.pickerDidDismiss` fires for the FIRST (rejected)
+  // sheet during this window — without this guard it resolves `pickerResult`
+  // as cancelled before the retry gets a chance to report its own outcome,
+  // so a successfully provisioned accessory gets reported to Dart as cancelled.
+  private var retryInFlight = false
 
   struct PickerError: Error { let message: String }
 
@@ -109,6 +115,9 @@ private final class Impl {
     // pending showPicker as "cancelled".
     switch event.eventType {
     case .pickerDidDismiss:
+      // Ignore the first sheet's dismissal while the Gen 4 retry is in flight —
+      // see `retryInFlight`'s doc comment.
+      guard !retryInFlight else { return }
       // If a picker was in flight and nothing got added, treat as cancelled. (If an
       // accessory WAS added, showPicker's completion handler already resolved it.)
       if let cb = pickerResult {
@@ -194,13 +203,25 @@ private final class Impl {
   private func present(_ items: [ASPickerDisplayItem], allowGen4Retry: Bool) {
     session.showPicker(for: items) { [weak self] error in
       guard let self = self else { return }
+      // The retry (if any) that led to THIS completion running is no longer
+      // in flight — whatever we resolve below is the actual outcome.
+      self.retryInFlight = false
       if let error = error {
         guard let cb = self.pickerResult else { return }
         let message = error.localizedDescription
-        let looksCancelled = message.lowercased().contains("cancel")
-        if allowGen4Retry, !looksCancelled, items.count > 1 {
+        // Prefer the typed error code over sniffing the localized message —
+        // a message that doesn't happen to contain "cancel" would otherwise
+        // incorrectly trigger a second picker on a real user cancellation.
+        let cancelled: Bool
+        if let askError = error as? ASError {
+          cancelled = askError.code == .userCancelled
+        } else {
+          cancelled = message.lowercased().contains("cancel")
+        }
+        if allowGen4Retry, !cancelled, items.count > 1 {
           NSLog("[ASK] picker rejected the %d-item descriptor list (%@) — "
                 + "retrying with the WHOOP 4.0 item only.", items.count, message)
+          self.retryInFlight = true
           self.present([items[0]], allowGen4Retry: false)
           return
         }
--- a/ios/Runner/AccessorySetup.swift
+++ b/ios/Runner/AccessorySetup.swift
@@ -26,13 +26,19 @@ import AccessorySetupKit
 ///   - `removeAll`          -> nil    (deprovision all — used on unpair)
 enum AccessorySetup {
   private static let channelName = "openstrap/accessory_setup"
-  // WHOOP GATT service UUIDs, one per generation (match GattProfile in Dart).
-  // `fileprivate` so the iOS-18 Impl below can read them. BOTH must also be
-  // listed in Info.plist under NSAccessorySetupBluetoothServices.
-  //   • gen4 ("Harvard", WHOOP 4)  — 6108…
-  //   • gen5 ("fd4b", WHOOP 5)     — fd4b…  (EXPERIMENTAL)
+  // WHOOP GATT service UUIDs (match GattProfile / kWhoopMemberUuid16 in Dart).
+  // `fileprivate` so the iOS-18 Impl below can read them. Every criterion used
+  // in an ASDiscoveryDescriptor must also be listed in Info.plist or iOS
+  // silently ignores it.
+  //   • gen4 ("Harvard", WHOOP 4)           — 6108… 128-bit vendor service
+  //   • gen5 ("fd4b", WHOOP 5.0 / MG)       — fd4b0001-cce1-… 128-bit vendor
+  //   • 16-bit SIG member UUID 0xFD4B       — what still fits a 31-byte AD
+  // The 16-bit form is NOT 0000FD4B-0000-1000-8000-00805F9B34FB; no band
+  // advertises that Bluetooth-base expansion.
   fileprivate static let whoopServiceUUIDGen4 = "61080001-8d6d-82b8-614a-1c8cb0f8dcc6"
   fileprivate static let whoopServiceUUIDGen5 = "fd4b0001-cce1-4033-93ce-002d5875f58a"
+  fileprivate static let whoopMemberUUID16 = "FD4B"
+  fileprivate static let nameSubstring = "WHOOP"
 
   static func register(messenger: FlutterBinaryMessenger) {
     let channel = FlutterMethodChannel(name: channelName, binaryMessenger: messenger)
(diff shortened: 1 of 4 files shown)


abdulsaheel:
thanks for digging into this one, but i can't take it as is — the 16-bit uuid isn't ours.

`0xfd4b` is assigned to samsung in the bluetooth sig member list. there's no whoop entry in there at all. our `fd4b0001-cce1-...` is a vendor-chosen prefix exactly like gen4's `6108`, same as `band.dart` says — only the 32-bit prefix changes between generations. nobody would read `0x6108` as a sig assignment either.

two things fall out of that:

**it breaks gen4 pairing near a samsung device.** `Guid('fd4b')` expands to `0000fd4b-0000-1000-8000-00805f9b34fb`, so a galaxy watch passes both the os `withServices` filter and `advertisementLooksLikeWhoop`. `scan()` takes the first match and stops, pairing feeds it into `pairWith()`, discovery hard-fails at "no whoop service". on android and ios <18 that's the actual pairing path. and that expansion is the same constant #237 already identified as "one wrong constant, zero results".

**the ask item can only ever match a samsung accessory.** `CBUUID(string: "FD4B")` is a different 128-bit value from `fd4b0001-cce1-...`, so the row labelled "whoop 5.0 / mg" lists galaxy buds and nothing else.

the name-only item can't work either — apple's `ASDiscoveryDescriptor` requires a service uuid or company id alongside `bluetoothNameSubstring`, so that list gets rejected. and the fallback then presents `[items[0]]` which is gen4 only, while main offers gen4 + the gen5 128-bit item. so a whoop 5 that does advertise `fd4b0001...` in its primary ad — the case main handles today — stops pairing under this.

also worth pushing back on the premise: apple scopes the overflow area to `CBPeripheralManager`, i.e. an ios device acting as peripheral. a third-party band's oversized uuid goes in the scan response, which corebluetooth merges into `advertisementData` normally.

smaller stuff:
- cancel is detected with `message.lowercased().contains("cancel")` — false in any non-english locale. `ASErrorCodeUserCancelled = 700` exists, use that.
- coderabbit's point about the retry racing `pickerDidDismiss` is right — a successful pair can get reported to dart as cancelled. worth fixing whatever else happens here.
- the tautological path assertion in the test (comparing a literal to itself after `_posix`) can go.

what i'd take right now: pull `kWhoopMemberUuid16` and both items using it, drop the name-only item, keep the `advertisementLooksLikeWhoop` extraction. that's a small clean refactor that changes no behaviour and i'll merge it.

if the scan-response case turns out to be real, the route is `bluetoothCompanyIdentifier` — it reads manufacturer data, which never gets hashed, and it satisfies apple's requirement on its own so it can legally carry the name substring too. but the company id has to come off a real advertisement, not a guess.

one more: there's a comment naming a capture tool. can you take that out and just say it's not confirmed on hardware? repo's public.

PR #260, opened 2026-08-19
gen5 link conformance: official bootstrap, a real burst count gate, command correlation

DropTabl:
gen5 link conformance: bootstrap order, a real burst count gate (gen5-scoped),
command correlation.

- **Bootstrap (gen5):** HELLO first during connect setup — identity, battery,
  charging, wear and the timestamp all land before any clock work; the clock
  decision takes hello's own timestamp, so GET_CLOCK becomes the fallback it
  is supposed to be. One SET_CLOCK per bootstrap, gated on ≥2 s of measured
  drift. The 600/500 ms registration delays, the final advertising-name read
  and the charging-only battery-pack follow-up are in their places.
- **Command correlation:** `CommandAwaiter` — observer registered before the
  write, both sequence and opcode must match, cancel on write failure,
  `failAll()` on teardown, and a serialized seq-zero fallback that refuses to
  guess when two requests share an opcode. Five hello failures reset the
  platform bond. Alarm arms are judged on the strap's correlated reply — a
  rejection never persists a phantom alarm.
- **Count gate (gen5 only):** expected = data + event packet counts; complete
  type-47/48/50/53/54/55 frames count once each, in arrival order, within the
  marker-to-marker window only; a short burst is refused with the two-byte
  failure result and the band re-offers; attempts are per-burst, the tally is
  frozen at the terminal, and the 15th consecutive failure sends one abort
  and latches terminal for the session. gen4 keeps its proven advisory-only
  flow untouched.
- **Honest v18 persistence:** hr gated on its own range, on-wrist and
  hr-valid never fabricated, skin temp abstains on the -50.00 °C sentinel —
  enforced at the backup-import seam too.
- Wake window at the 180 s / 7200 s Smart Alarm values.

Compiles against protocol #31 (repin to the upstream SHA as the final commit
once it merges — merge order: #31, repin, then this).

## Summary by CodeRabbit

- **New Features**
  - Improved Bluetooth command handling with response matching, timeouts, identity checks, and safer startup validation.
  - Added support for burst-count data during active historical synchronization.
  - Added battery-pack usability and clock-drift checks during connection setup.

- **Bug Fixes**
  - Historical synchronization now stops cleanly when the connection cannot progress.
  - Alarm refusals and transport failures no longer save invalid alarm states.
  - Removed unsupported wear, heart-rate validity, and temperature values from imported health data.

- **Improvements**
  - Updated protocol support for newer Gen5 records without changing existing derived results.


DropTabl:
@coderabbitai review

abdulsaheel:
the command-correlation layer is good and i want it. the burst count gate isn't ready.

**the gate can trim flash for records we never got.** three separate ways:

1. `consecutiveValidationFailures` resets only on success — not in `beginBurst()`. so burst a fails 3 times, burst b arrives, and b's **first** attempt gets `burstCountSlack(3) == 2`. short by 2 → passes → `commit(token)` → band trims two frames we never counted. that directly contradicts the rule the code says it implements ("the first three attempts demand every frame"). the 15-attempt terminal counts across different bursts too, so 15 separate bursts each failing once latches `historyStuck` while the log says "still short after 15 attempts", which isn't what happened. one line in `beginBurst()`.

2. on a re-offered HISTORY_END there's no reset — right, since the band re-offers the marker without resending frames. but event/console/puffin frames arriving during the ~37s retry window now go through `immediateAndCount` and bump the same counter. so a burst short by 3 data frames gets pushed over by three unrelated console logs, then commits and acks. gen5 console traffic is chatty enough for that to happen. the gate compares a total, so it can't tell "the missing frame came back" from "a different frame arrived". freeze the tally at HISTORY_END, or only count members that arrived before the terminal.

3. **there's no `isGen5` guard anywhere on this path.** the comment this pr deletes is my own contrary evidence — field data showed the gap between expected and actual varies run to run with no fixed offset, and the old hard-fail behaviour gave zero sync progress with "last data" frozen indefinitely. the membership rules here came off gen5 work and were tested on a whoop 5. short by ≥3 on gen4 now means 15 failures → abort → `historyStuck` latched for the connection. records already banked survive, so it's a stall not immediate loss, but on a full band a stall becomes loss. gate it on gen5 or get gen4 evidence.

same path: 3+ crc-failed frames terminating history for the whole connection is rough on a marginal link, since crc failures never reach `receivedTrafficCount` and slack caps at 2.

**provenance.** there's a lot of it — comments citing source documents by name, "official client" references, and one that ships in a **runtime log line** users can see. plus the pr body. state the fact, never how it was learned. repo's public, force-pushed shas persist, github won't delete a pr. has to be sorted before merge.

**merge order.** doesn't compile against the pinned protocol — 29 analyzer errors, missing symbols from #31. you say so in the body and it's real: #31 lands, edge repins, then this. also `pubspec.lock` is committed with a local override baked in (`path: "../protocol"`) — that's a `pubspec_overrides.yaml` artifact, revert it. the repin belongs in `pubspec.yaml`.

smaller:
- `ConditionalWakePolicy.fireNow` doesn't set `windowOpen`, so the next tick issues `openWindow` after the user's already been woken. coderabbit flagged this too, it's right.
- `historyStuck` returns before every sub-type including `historyComplete`, so `onComplete()` can never fire once latched and `awaitComplete()` waiters run to timeout.
- `_readClock` now needs seq correlation on gen4 and i have no gen4 capture with `cmd_response` to check the echo against. if gen4 doesn't echo the request seq you get a 3s stall on every connect. the seq-zero fallback probably covers it, but worth one connect log.
- event/console frames now re-arm the idle watchdog mid-offload, so chatter can keep a genuinely stalled offload alive past 60s.
- `ConditionalWakePolicy`, `runStoredAlarm` and the `StrapConditionReport` fields have no production caller — that's smart-alarm machinery in a pr called "link conformance". split it out.
- `BatteryPackInfoGate.usable` only rejects the sentinel in `identifier`, so an empty id with a `00:00:...` name passes.
- `BurstShortfallGate` is dead now — delete it or say why it stays.
- `_mergeFromDbFile` still includes `decoded_onehz`, so a pre-v46 backup re-imports exactly the rows v46 retires. coderabbit's right about that one.

what's genuinely good, so it doesn't get lost in the above: `CommandAwaiter`/`_sendAwaited` is well built — observer registered before the write, cancel on write failure, both seq and opcode must match, `failAll()` on teardown, and the seq-zero fallback refuses rather than guesses when two requests share an opcode. i traced every caller, nothing leaks a completer. judging the alarm arm on the strap's reply so a rejection never persists a phantom alarm is right, and so is letting an *unanswered* arm still succeed. `skinTempCOrNull` abstaining on the -50c sentinel instead of banking a number 70c below any wrist is exactly the rule. and gen4's connect chain is structurally unchanged — i traced it end to end. dropping the bogus gen5-only opcodes back to 10/11 is a good catch.

DropTabl:
Round one of fixes is up as one additive commit (6ccaae2); the smart-alarm machinery is out and the count gate got all three of your data-loss scenarios closed. Point by point:

**Gate trims flash for records we never got:**
1. `beginBurst()` now resets `consecutiveValidationFailures` — a new HISTORY_START is a fresh cycle at slack zero, and 15 different bursts each failing once can no longer latch Stuck under a log claiming one burst failed 15 times. Marker-only re-offers (no START) still accumulate, which per your description is what the band actually sends after a failure result. Pinned by "a new HISTORY_START starts a fresh validation cycle".
2. The tally freezes at HISTORY_END (`closeBurstTally`): event/console members arriving during the re-offer window count into no burst, so chatter can't push a short tally over the line. Pinned by "chatter after HISTORY_END cannot push a short burst over the line".
3. The gate is enforced on gen5 only. gen4 keeps the advisory-only behaviour it always had — mismatch logged (`ADVISORY, gen4`) and written to the ledger, ACK unchanged — until a gen4 capture pins its count semantics. Pinned by "gen4 keeps the advisory-only count behaviour".

On the CRC point: still real on gen5 — a marginal link corrupting ≥3 frames per delivery can walk to the 15-attempt boundary. Refusing stays the data-safe direction (every re-delivery is another chance at the corrupt frame; an ACK deletes it), and the boundary keeps it from looping forever. I'd rather revisit it as part of making `Stuck` task-scoped instead of connection-scoped, which I want to bring as its own PR after this lands.

**Provenance** — scrubbed from the whole diff: comments, test names, the one runtime log line, and this PR's body (edited in place). Facts stay, sources don't.

**Smaller items:**
- `historyStuck` now lets `HISTORY_COMPLETE` through (it ACKs nothing), so `onComplete()` fires and `awaitComplete()` waiters stop running out their timeout. Pinned.
- event/console frames no longer re-arm the idle watchdog — only real drain progress does. Pinned ("console chatter does not keep a stalled offload alive").
- `_readClock` logs one line when a reply matches via the seq-zero fallback, so a gen4 field connect answers your echo question before anything gates harder on it.
- `ConditionalWakePolicy`, `runStoredAlarm` and the condition-report fields are OUT of this PR (event 29 still gets one stateless log line). They come back as a smart-alarm PR, with the `fireNow`/`windowOpen` fix you and coderabbit flagged.
- `BatteryPackInfoGate` rejects the sentinel in the name field too (your exact case: empty id + sentinel name). A real address with a junk name still passes — the address alone identifies the pack.
- `BurstShortfallGate` deleted.
- The v46 data rule now also applies at the backup-import seam, with a regression test importing a pre-v46 source and asserting on_wrist/hr_valid/the -50 °C sentinel all land NULL. I did NOT hard-null the shared `_queueDecodedOneHz` write: those columns are deliberately kept writable for a source that can honestly supply them, every current producer already passes null, and the import seam was the one ingress that could carry disproven values.
- coderabbit's `register()`-leak finding: verified not live — both `_sendAwaited` return paths evaluate `pending.response` (which arms the timeout) and the write-fail path cancels, matching your own trace. The lazy arm stays as the documented sharp edge; happy to switch to arm-at-registration if you'd rather remove the edge than document it.
- The alarm exception is neutral now ("Alarm not set") — null covers both a failed write and an explicit refusal, and the engine log says which.
- `pubspec.lock` restored to main's git-pinned form; the repin lands in `pubspec.yaml` as the final commit once #31 merges, in your stated order (and after the 146/147 question over there is settled).

One heads-up: your protocol-side v25 change (accelG now empty) reached this branch through the local override, so `v25_refusal_test` now pins the upstream refusal instead of asserting protocol "still hands us the vector". Same seams, same records dropped.

DropTabl:
Round two. #31 is merged, so this is repinned and out of draft. Suite is green: **2,910 passed, 422 skipped, 0 failures**, analyzer clean.

Four commits on top of `6ccaae2`:

**`e524c34` — merge main.** Two conflicts. `ble_engine`: kept the `!_burstTallyClosed` guard, adopted your three-argument signature. `v25_refusal_test`: took main's version, since both sides had independently reached the same conclusion and main's also asserts on a second fixture.

**One thing in that merge is new code you have not reviewed, and it is not cosmetic.** Moving the revision derivation up to the caller left this branch with two call sites; git flagged one. The other is the raw-archive path added for the count gate. `historicalPacketCount` sums the per-revision buckets and excludes `_unknownCount`, so passing `-1` there would have left every burst carrying a raw archive permanently short — the same data-loss class as the tally fixes in the last round, arriving through a merge rather than a decision. It reads the revision off `inner[1]` exactly as the callee used to, and passes `-1` only for a frame too short to have one.

**`2e43bb1` — you were right about `Stuck`, and I was wrong to defer it.** I said I would rather revisit it in a follow-up. That was a bad answer: the latch is new in this PR, not something inherited, so its blast radius is mine to fix here.

The refusal stays — refusing a short burst rather than ACKing it is the data-safe direction, since every re-delivery is another chance at the corrupt frame and an ACK deletes it. The latch also earns its keep against the ~2.5 s re-offer storm. What I could not defend is the scope. Continuation after `Stuck` comes from a later connection, a scheduler tick or an explicit trigger, and a session-scoped latch refused the last two outright — stricter than the behaviour it models, and exactly your point about a marginal link costing every later drain.

So it is windowed now: `kHistoryStuckCooldown` = 2 min, outlasting both the re-offer storm and the 60 s idle watchdog. Inside the window nothing changes. After it, a new trigger gets a fresh cycle; the band still holds its checkpoint. Every decision site reads one windowed getter — including `historyStuckThisSession`, which the backfill loop uses to mirror the engine's refusal and would otherwise have kept breaking the loop after the engine had resumed accepting triggers. The raw latch stays in diagnostics as `history_stuck`.

**`8aa0f64` — repin to `b7990e1`**, protocol main's merge commit rather than the PR-branch head, per this file's own rule about orphaned SHAs. Lock regenerated with no local override in it: git source, url, resolved-ref. You were right that the `path: "../protocol"` in the committed lock was a `pubspec_overrides.yaml` artifact.

**`40fcb30` — the sibling-pin guard caught the repin**, which is what it is for. Holding `kAlgoVersion` at 76 rather than bumping, and this one is checkable rather than argued: diff `c761f29..b7990e1` and the gen4 record decoder (`lib/src/records.dart`) is untouched, as is every gen4 line in the package export. Only the gen5 surface moved. So for anyone on a gen4 strap every number out of the package is byte-identical across the repin, and a bump would invalidate every stored day to recompute the same answers. The gen5 records it adds are new — no released build could decode them, so no stored day at v76 came from one.

**Still open, your call:**

- **`register()`.** You are right that documenting the edge is weaker than removing it. "Call `register()` before the write" is enforced by the API's shape; "always evaluate `.response`" is only enforced by every current caller happening to do it — a property of the callers, not the type. I have not made the change, because the one side effect worth naming is that arming at registration starts the clock before the write completes, shrinking the window by the write latency. Under 1% against a 5 s timeout, and arguably the more correct deadline. Say the word and it lands here rather than as a follow-up.
- **CRC and `Stuck` scoping.** The cooldown bounds the damage, but making `Stuck` task-scoped instead of connection-scoped is still the better shape. I would still rather bring that as its own PR than widen this one — happy to be overruled.

DropTabl:
CodeRabbit pass — all seven findings accounted for. Suite green: **2,910 passed, 422 skipped, 0 failures**, analyzer clean.

**Fixed in `c67d353`:**

**`register` can leak entries that permanently break the sequence-zero fallback** (🟠 Major) — real, and the leak was worse than memory. The expiry armed only when a caller read `PendingCommand.response`, so an entry registered whose write path returned without awaiting and without `cancel()` stayed in `_pending` for the life of the connection. `deliver` then refused every later sequence-zero fallback for that opcode, because the stale entry made `sameOpcode` ambiguous — valid replies came back unmatched and their callers timed out.

The lazy arm is now an explicit one-shot `Timer` started in `register`. `response` still arms idempotently for callers that reach it first, and `cancel`/`_complete` cancel it, so the timeout is still applied exactly once with no automatic resend. This is the arm-at-registration option from the last round: you were right that removing the edge beats documenting it. The one side effect worth naming is that the clock now starts fractionally before the write returns — a few ms out of a multi-second window, in exchange for the invariant that nothing can outlive its timeout.

**Restore complete API documentation** (🟡 Minor) — real, and my fault from the provenance scrub. Three doc comments lost their subject and read as fragments: `The identity half of as an OBSERVATION.`, `The bootstrap clock gate from .`, and an orphaned `.`. Rewritten as complete neutral descriptions that say what the gate or observation represents without saying where it came from. I swept every changed file for the same pattern — those three were the only ones.

**Already addressed, verified against current code:**

- **Set `windowOpen` when `fireNow` is returned** — stale. `ConditionalWakePolicy` and the whole smart-alarm surface came out of this PR in `6ccaae2`; there is no `fireNow` path left here. It comes back with that fix already applied when the smart-alarm PR lands.
- **Apply the all-zero sentinel to `name`, not only to `identifier`** — done in `6ccaae2`. `BatteryPackInfoGate.usable` now returns false for an empty identifier with a sentinel name; a real address with a junk name still passes, since the address alone identifies the pack.
- **Use a neutral error message for both null outcomes** — done in `6ccaae2`. It is `Exception('Alarm not set')`; null covers both a failed write and an explicit refusal, and the engine log says which.
- **Apply the retirement rule to every decoded-row ingress** — the import seam is fixed, which this batch's own follow-up confirms. The shared `_queueDecodedOneHz` write is deliberately left writable: those columns exist for a source that can honestly supply them, every current producer already passes null, and the import seam was the one ingress that could carry disproven values in.
- **Regression test still worth adding** — it is already there, in `test/db_paged_import_export_test.dart` (`importFromDbFile applies the v46 data rule at the seam`). It builds a pre-v46 `decoded_onehz`, imports it, and asserts `on_wrist` and `hr_valid` land NULL and the `-50.00 °C` sentinel clears, while a real `30.57` reading survives untouched. It was added in `6ccaae2`, so this batch's diff would not have shown it.

That clears everything outstanding from both review passes. The only item I have left deliberately open is making `Stuck` task-scoped rather than connection-scoped — the cooldown in `2e43bb1` bounds the damage, but the better shape is still its own PR, and I would rather not widen this one further. Happy to be overruled.

Issue #186, opened 2026-08-04
Smart Alarm does not trigger when scheduled


gerageragarza-prog:
### What happened?

I scheduled a Smart Alarm to test the feature.

### Steps to reproduce

1. Open Edge.
2. Go to Smart Alarm.
3. Schedule a Smart Alarm.
4. Wait until the scheduled time.
5. The band does not vibrate.

### App version

v0.9.23

### Platform

iOS (TestFlight)

### Phone model and OS version

26.5.2

### Which band?

WHOOP 5.0 / MG (experimental — expect breakage)

### Screenshots

<img width="1206" height="516" alt="Image" src="https://github.com/user-attachments/assets/324d38ae-73b2-46cd-a809-80cc039c7018" />

### Quick checks

- [x] The official WHOOP app is not also connected to this band
- [x] I've searched existing issues for this

gerageragarza-prog:
I'm testing every aspect of the app to help calibrate and fine-tune each feature.

abdulsaheel:
is it for whoop 5 ?

gerageragarza-prog:
Yes, it's for WHOOP 5.0/MG.

I did some research with Claude and came across this. I hope it helps—it's part of NOOP, and it might give us some useful clues.


PROTOCOL.md completo (la sección de comandos de alarma es la §6 "CommandNumber"): https://github.com/ryanbr/noop/blob/main/docs/PROTOCOL.md

Commands.swift (donde está setAlarmPayload y el builder de frames): https://github.com/ryanbr/noop/blob/main/Strand/BLE/Commands.swift

Framing.swift (CRC8/CRC16/CRC32 y verificación de frames): https://github.com/ryanbr/noop/blob/main/Packages/WhoopProtocol/Sources/WhoopProtocol/Framing.swift

BLEManager.swift (bonding y transporte CoreBluetooth): https://github.com/ryanbr/noop/blob/main/Strand/BLE/BLEManager.swift

AutomationsView.swift (la UI de "Smart alarm"): https://github.com/ryanbr/noop/blob/main/Strand/Screens/AutomationsView.swift

Brackyt:
I fixed this here https://github.com/OpenStrap/edge/pull/194

Brackyt:
Btw I think noop doesn't work too, last time I tried they had all the same issues I fixed here in multiple PRs

gerageragarza-prog:
That's awesome! Thank you so much for your contribution, my friend!

[...]

PR #265, opened 2026-08-20
fix(ble): arm gen4 wake alarm with the rev-1 9-byte form the firmware executes

### LINKED ISSUE #119: Smart alarm never fires (strap never buzzes at wake time)
## Summary
The smart / cycle-aligned wake alarm never actually fires — the band does **not** buzz at the scheduled wake time, even when an alarm is legitimately set.

## Steps to reproduce
1. Set a band alarm via the Sleep coach ("Set band alarm for HH:MM").
2. Wait until the scheduled wake time.

## Observed
The strap never buzzes; no wake alarm fires.

## Expected
The band vibrates at the scheduled wake time.

## Notes
Likely in the BLE set-alarm delivery/trigger path (command build → BLE write → strap arming, and time encoding). May be related to the strap confirmation never returning — see #118. Root cause under investigation; opcode/payload to be cross-checked against the OpenStrap protocol library's alarm support.

## Screenshots
_The Sleep page "Set band alarm" UI is shown in #118; this issue concerns the delivery behaviour rather than the UI._

Pablodvs:
# Smart alarm never fires on WHOOP 4.0 — root cause and fix

**Issue:** #119 (gen4, closed by #122 without on-device verification — the drift
fix it shipped never activates and did not address the cause). Related: #186
(same symptom reported on WHOOP 5/MG; the gen5 arm path is NOT changed here).

## Symptom

A scheduled smart alarm is accepted and confirmed by the band — the app shows
it armed — but the band never vibrates at the wake time. The test buzz
(RUN_ALARM) and Find-my-band haptics always work.

## What was actually happening

Edge armed gen4 with the **20-byte `0x04` "rich" SET_ALARM_TIME form**. The
gen4 firmware **stores, echoes and confirms** that form exactly like a real
arm (exact epoch in the `0x42` response and in the ALARM_SET event-56 payload)
— **but its scheduler never executes it**. Across 1.07M lines of sync logs
covering 8+ armed alarms on a real WHOOP 4.0 there is not a single
STRAP_DRIVEN_ALARM_EXECUTED (event 57) and no haptics at any armed target.

Everything else was ruled out on hardware: the band's RTC is wall-correct to
the second (raw GET_CLOCK readbacks match wall; SET_CLOCK latches exactly),
the armed epoch is byte-exact, the band was on-wrist at every target, MTU/
writes/link were healthy.

Two things had masked this for months:

1. The band re-sends stored event frames during every history sync through the
   same live callback, and the timestamp-blind confirmation machine treated
   them as live — a replayed test-buzz EXECUTED(58) landing at the wake minute
   looked exactly like "the alarm fired silently". (Known defect, not fixed in
   this change.)
2. The pinned protocol decoder misparsed GET_CLOCK, so `correlated=false` on
   every arm pointed suspicion at clock drift (#122's theory). Upstream fixed
   the parse in `openstrap_protocol` `95d8ab8`; this repo picked it up with the
   0.9.27 repin — and the alarm still never fired, because drift was never the
   cause.

## The fix

Arm gen4 with the **REV-1 9-byte form** instead:

```
[0x01][epoch u32 LE][subsec u16 LE][haptic-mode u16 LE = 0]
```

This is what the **official WHOOP app sends** — btsnoop wire capture of the
official app arming a real 4.0 (noop PR #535; the capture's frame is pinned
byte-for-byte in `test/alarm_test.dart`). The trailing haptic-mode u16 is the
entire difference from the old known-silent 7-byte form.

Changes:

- `lib/ble/ble_state.dart` — new `AlarmPayloads.rev1()`, delegating to
  `openstrap_protocol`'s `alarmRev1Payload` (added there alongside
  `cmdSetAlarmRev1` for OpenStrap/protocol#32, so the wire format has
  exactly one home); `setPayloadForBand` now returns it for gen4. The gen5
  branch (rich 21-byte slot-1 body from #194) is unchanged.
  `rich()`/`simple()` stay as documented reference-only forms.
- `lib/ble/ble_engine.dart` — `setAlarm` docs + log line (`rev1 9B`); the
  drift-frame shift is kept (correct for a genuinely offset RTC).
- `test/alarm_test.dart` / `test/gen5_wiring_test.dart` — rev-1 layout +
  official-app wire-capture vectors; gen4 arm-body expectations updated.

Depends on OpenStrap/protocol#33: the pinned `openstrap_protocol` SHA
must contain `alarmRev1Payload` (pin bump left to the release-time bump,
so CI here stays red until then).

## Verification

A/B experiment over BLE against the affected band (2026-08-19), armed via a
reference client, band otherwise untouched:

- rich 20B form: latches + confirms (event 56), **never executes** — as always.
- rev-1 9B form: latches + confirms, and at the armed second the band fired
  **autonomously**: `HAPTICS_FIRED (60)`, `STRAP_DRIVEN_ALARM_EXECUTED (57)`
  and the one-shot auto-disable `(59)`, all stamped at the target epoch; ~24 s
  buzz, felt on wrist. First event-57 this band has ever emitted.

App-level: this build (0.9.27+59 + fix) installed on the affected phone;
`flutter analyze` clean; 24/24 alarm tests pass.

Also learned on hardware, relevant to reviewers: alarm lifecycle events
(56/57/59/60) are delivered via the band's **history stream on the next sync**,
not necessarily live — so an arm may show "waiting for the strap to confirm"
until the next sync, and the fired events arrive with the sync after the wake.

## Follow-ups (not in this change)

- Upstream the gen4 form correction to `OpenStrap/protocol` (its docs still
  call the rich form "the one that actually fires").
- Timestamp-gate `AlarmConfirmation` so drain-replayed 56/58/59 events cannot
  fake confirm/fired/cleared transitions (this is what closed #119 wrongly and
  once wiped an armed alarm a minute before its wake time).
- Re-test WHOOP 5/MG (#186): noop marks the gen5 rich form's actual firing as
  unverified; a gen5 equivalent of this A/B experiment would settle it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved alarm scheduling for WHOOP 4 devices by using the verified REV-1 alarm format.
  * Retained rich alarm support for WHOOP 5 devices, including slot and crescendo settings.
  * Clarified handling and logging of supported and reference-only alarm formats for easier troubleshooting.
  * Improved alarm status reporting by distinguishing executable formats from acknowledged but non-triggering formats.

### DIFF
--- a/lib/ble/ble_engine.dart
+++ b/lib/ble/ble_engine.dart
@@ -5600,13 +5600,15 @@ class BleEngine {
   ///   [5..7]   u16 subsec  LE    (millis % 1000) * 32768 ~/ 1000 (1/32768 s units)
   ///   [7..9]   u16 haptic-mode   0 = the stock wake buzz
   /// ```
-  /// This is what the official WHOOP app sends (btsnoop wire capture) and it
-  /// is proven on our own band (2026-08-19): armed over BLE, the band fired
-  /// autonomously at the armed second (HAPTICS_FIRED 60 + STRAP_DRIVEN_ALARM_
-  /// EXECUTED 57, then auto-disable 59). The rich 0x04 form previously armed
-  /// here is stored + confirmed (event 56) but NEVER executed on gen4 — the
-  /// silent-alarm root cause. The 7-byte short form ([setAlarmSimple]) is
-  /// rev-1 minus the haptic-mode u16: ACKed, never fires.
+  /// This is what the official WHOOP app sends (btsnoop wire capture), and on
+  /// our band (fw 41.17.4, 2026-08-19/20) it fired autonomously at the armed
+  /// second (HAPTICS_FIRED 60 + STRAP_DRIVEN_ALARM_EXECUTED 57, then
+  /// auto-disable 59) while the rich 0x04 form previously armed here latched
+  /// (event 56) without executing. Execution of the rich form is
+  /// firmware-dependent — at least one other WHOOP 4 executes it (see
+  /// [AlarmPayloads]). The 7-byte short form ([setAlarmSimple]) is rev-1
+  /// minus the haptic-mode u16 — the same bytes on the wire once padded, not
+  /// a distinct form.
   ///
   /// WHOOP 5 — the rich 21-byte slot-1 body, unchanged (#194; index 0 is
   /// rejected with `arm info is invalid, error 0xb`).
@@ -5649,8 +5651,9 @@ class BleEngine {
     // latched / drift) the raw wall epoch fires at the wrong strap-time — or
     // never (a raw wall epoch is decades ahead of a strap clock still near its
     // factory epoch). (Historical note: drift was once blamed for the gen4
-    // silent alarm; the real cause was the payload form — see the doc above.
-    // The shift stays: it is correct for a genuinely offset RTC.) Fall back to
+    // silent alarm, and later the payload form — which held on fw 41.17.4 but
+    // is firmware-dependent, see the doc above. The shift stays either way:
+    // it is correct for a genuinely offset RTC.) Fall back to
     // the raw epoch when we have no correlation yet (e.g. just after a
     // reconnect, before this session's GET_CLOCK reply). Frame conversion +
     // generation dispatch live in the pure [AlarmPayloads]; the gen4 rev-1
@@ -5705,13 +5708,14 @@ class BleEngine {
 
   /// Time-only alarm (SET_ALARM_TIME = 0x42), SHORT 7-byte form:
   /// `[0x01][u32 epoch-sec LE][u16 subsec LE]`. Kept for diagnostics/parity —
-  /// the band ACKs it but never fires it (it is the rev-1 form minus the
-  /// trailing haptic-mode u16, and those two bytes are what make the firmware
-  /// execute the alarm). Use [setAlarm].
+  /// it is the rev-1 form minus the trailing haptic-mode u16 and pads to the
+  /// identical frame when that u16 is 0, so it is not a distinct wire form.
+  /// Use [setAlarm], which also converts to the strap's RTC frame; this sends
+  /// the raw epoch.
   Future<void> setAlarmSimple(DateTime when) async {
     await _send(Cmd.setAlarmTime, AlarmPayloads.simple(when));
-    _log('SET_ALARM_TIME (simple 7B) → sec=${when.millisecondsSinceEpoch ~/ 1000} '
-        '(ACKs but will not fire)');
+    _log('SET_ALARM_TIME (simple 7B) → '
+        'sec=${when.millisecondsSinceEpoch ~/ 1000}');
   }
 
   /// Read the armed alarm back. Body is band-specific (see
--- a/lib/ble/ble_engine.dart
+++ b/lib/ble/ble_engine.dart
@@ -5590,19 +5590,30 @@ class BleEngine {
   /// frames queued ahead of the response, not to be a plausible steady state.
   static const Duration _clockReadTimeout = Duration(seconds: 3);
 
-  /// On-device wake alarm (SET_ALARM_TIME = 0x42) — the RICH 20-byte form that
-  /// actually FIRES:
+  /// On-device wake alarm (SET_ALARM_TIME = 0x42), band-generation aware via
+  /// [AlarmPayloads.setPayloadForBand]:
+  ///
+  /// WHOOP 4 — the REV-1 9-byte form the firmware actually EXECUTES:
   /// ```
-  ///   [0]      0x04              rich-form marker
-  ///   [1]      u8  index         alarm slot (gen4: 0; gen5: 1)
-  ///   [2..6]   u32 epoch-sec LE  the wake time
-  ///   [6..8]   u16 subsec  LE    (millis % 1000) * 32768 ~/ 1000 (1/32768 s units)
-  ///   [8..20]  12-byte haptic pattern (see [AlarmPayloads.defaultHaptics])
+  ///   [0]      0x01              rev-1 form marker
+  ///   [1..5]   u32 epoch-sec LE  the wake time
+  ///   [5..7]   u16 subsec  LE    (millis % 1000) * 32768 ~/ 1000 (1/32768 s units)
+  ///   [7..9]   u16 haptic-mode   0 = the stock wake buzz
   /// ```
-  /// WHOOP 5 requires slot index 1: index 0 is
-  /// rejected with `arm info is invalid, error 0xb`. The short 7-byte
-  /// time-only form ([setAlarmSimple]) is ACKed but never buzzes. The strap
-  /// confirms via event 56 and reports firing via 57/58 + 60.
+  /// This is what the official WHOOP app sends (btsnoop wire capture) and it
+  /// is proven on our own band (2026-08-19): armed over BLE, the band fired
+  /// autonomously at the armed second (HAPTICS_FIRED 60 + STRAP_DRIVEN_ALARM_
+  /// EXECUTED 57, then auto-disable 59). The rich 0x04 form previously armed
+  /// here is stored + confirmed (event 56) but NEVER executed on gen4 — the
+  /// silent-alarm root cause. The 7-byte short form ([setAlarmSimple]) is
+  /// rev-1 minus the haptic-mode u16: ACKed, never fires.
+  ///
+  /// WHOOP 5 — the rich 21-byte slot-1 body, unchanged (#194; index 0 is
+  /// rejected with `arm info is invalid, error 0xb`).
+  ///
+  /// The strap confirms via event 56 and reports firing via 57/58 + 60 —
+  /// delivered through the band's history stream (typically the NEXT sync),
+  /// not necessarily live.
   ///
   /// Returns the wall-clock instant armed, or null when the strap did not take
   /// the alarm — so the caller never persists a phantom alarm. Null means one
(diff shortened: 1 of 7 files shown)


abdulsaheel:
thanks for chasing this, and the doc corrections in here are right. but i can't take the behaviour change — the mechanism it rests on isn't there, and my own band contradicts the premise.

**the two forms are byte-identical on the wire.** `pad4` zero-pads inner to a multiple of 4 and the unpadded length is never transmitted. short form: inner `[0x23, seq, 0x42]` + 7 = 10, padded to 12 with two `0x00`. rev-1: 3 + 9 = 12, last two bytes `hapticMode = 0`. i built both:

```
simple: aa 10 00 57 23 01 42 01 30 d5 35 6a 00 00 00 00 14 55 c3 a3
rev1  : aa 10 00 57 23 01 42 01 30 d5 35 6a 00 00 00 00 14 55 c3 a3
```

same frame. the radio can't tell them apart, so "those two bytes are the whole difference" can't be why one fires and the other doesn't. both repos would ship a self-contradiction after this — `cmdSetAlarmSimple` saying "acked, never buzzes" next to `cmdSetAlarmRev1` "the form that fires", for identical bytes.

which cuts the other way too: if your a/b result is real, it also shows the **short form fires**. the "acks but never buzzes" line has been in both repos unsupported for a while and should just be deleted, not re-explained.

**and the rich form does fire on my 4.0.** from my own export, 2026-08-11: event 56 `strapDrivenAlarmSet` with a body carrying the armed epoch and the haptic block `47,152,0,0,0,0,0,0,0,0,7,30` — that's the rich 20-byte form. then 57 `strapDrivenAlarmExecuted` at the byte-exact armed second, and haptics terminated at target **+30s**, matching `durationSeconds = 30` in that same block. you attribute ~24s to the rev-1 stock buzz, so the 30s is independent evidence the firmware played the rich waveform.

so the honest claim is "rich doesn't fire on that band", not "gen4 never executes it". my guess at the real discriminator is firmware version — neither the pr nor #32 mentions it. can you post `getVersionInfo` off your band? i'll do mine.

**one thing i want fixed regardless of who's right about the form.** this makes 57/59/60 real for the first time, and `_handleAlarmEvent(int id, int ts)` takes a timestamp and never uses it — it calls `onEvent(id, DateTime.now())`. those come through the history-replay path. so a replayed 57 wipes `_savedAlarm`, `device.alarmEpoch` and the pref, and pops a phantom "your alarm went off". it already half-happens: my `notif_fired` has an `alarm_fired` key stamped 91 minutes after the alarm actually fired, i.e. at sync time. the timestamp gate is listed as a follow-up — it has to ship with this, not after.

if the rich form turns out not to fire for you, the minimal change is pointing `setPayloadForBand`'s gen4 branch at the existing `simple(when)` plus the doc fixes. that drops the new builder entirely and makes the edge side independently mergeable — it currently can't compile against the pinned protocol and the pr doesn't include the repin, so as written it can't be both mergeable and green.

smaller:
- `hapticMode` names a meaning nothing supports — the only established fact is that zeros go there. call it `reserved` if a builder survives.
- coderabbit's right that `cmdSetAlarmRev1` shouldn't accept gen5 at all, and that the timestamp should be range-checked before truncation. odd that the new function validates `hapticMode` but not `when`.
- `app_state.dart:3672` still says the engine computes real sub-seconds "for the rich 20-byte firing form" — that one's missed by the doc sweep.

**provenance**: there are comments and test names across both prs that say how the bytes were obtained, two of which also name the upstream project, and #32 publishes a band serial. can you strip those to just the vector — "epoch 1781912880 → `01 30 d5 35 6a 00 00 00 00`" says everything useful. pr and issue bodies are editable, worth doing before any of this merges.

what's right and i want kept: deleting "the rich form is THE form that actually fires" as an unqualified claim, "a GET_ALARM readback proves only that a body was STORED", "56 proves latch, not execution", and removing the fabricated `idx` that was reading an epoch byte as a slot id and inventing `-1` otherwise. gen5 is genuinely untouched in both directions, which i checked.

(same comment on protocol#33 — they only make sense as a pair.)

Pablodvs:
> thanks for chasing this, and the doc corrections in here are right. but i can't take the behaviour change — the mechanism it rests on isn't there, and my own band contradicts the premise.
> 
> **the two forms are byte-identical on the wire.** `pad4` zero-pads inner to a multiple of 4 and the unpadded length is never transmitted. short form: inner `[0x23, seq, 0x42]` + 7 = 10, padded to 12 with two `0x00`. rev-1: 3 + 9 = 12, last two bytes `hapticMode = 0`. i built both:
> 
> ```
> simple: aa 10 00 57 23 01 42 01 30 d5 35 6a 00 00 00 00 14 55 c3 a3
> rev1  : aa 10 00 57 23 01 42 01 30 d5 35 6a 00 00 00 00 14 55 c3 a3
> ```
> 
> same frame. the radio can't tell them apart, so "those two bytes are the whole difference" can't be why one fires and the other doesn't. both repos would ship a self-contradiction after this — `cmdSetAlarmSimple` saying "acked, never buzzes" next to `cmdSetAlarmRev1` "the form that fires", for identical bytes.
> 
> which cuts the other way too: if your a/b result is real, it also shows the **short form fires**. the "acks but never buzzes" line has been in both repos unsupported for a while and should just be deleted, not re-explained.
> 
> **and the rich form does fire on my 4.0.** from my own export, 2026-08-11: event 56 `strapDrivenAlarmSet` with a body carrying the armed epoch and the haptic block `47,152,0,0,0,0,0,0,0,0,7,30` — that's the rich 20-byte form. then 57 `strapDrivenAlarmExecuted` at the byte-exact armed second, and haptics terminated at target **+30s**, matching `durationSeconds = 30` in that same block. you attribute ~24s to the rev-1 stock buzz, so the 30s is independent evidence the firmware played the rich waveform.
> 
> so the honest claim is "rich doesn't fire on that band", not "gen4 never executes it". my guess at the real discriminator is firmware version — neither the pr nor #32 mentions it. can you post `getVersionInfo` off your band? i'll do mine.
> 
> **one thing i want fixed regardless of who's right about the form.** this makes 57/59/60 real for the first time, and `_handleAlarmEvent(int id, int ts)` takes a timestamp and never uses it — it calls `onEvent(id, DateTime.now())`. those come through the history-replay path. so a replayed 57 wipes `_savedAlarm`, `device.alarmEpoch` and the pref, and pops a phantom "your alarm went off". it already half-happens: my `notif_fired` has an `alarm_fired` key stamped 91 minutes after the alarm actually fired, i.e. at sync time. the timestamp gate is listed as a follow-up — it has to ship with this, not after.
> 
> if the rich form turns out not to fire for you, the minimal change is pointing `setPayloadForBand`'s gen4 branch at the existing `simple(when)` plus the doc fixes. that drops the new builder entirely and makes the edge side independently mergeable — it currently can't compile against the pinned protocol and the pr doesn't include the repin, so as written it can't be both mergeable and green.
> 
> smaller:
> 
> * `hapticMode` names a meaning nothing supports — the only established fact is that zeros go there. call it `reserved` if a builder survives.
> * coderabbit's right that `cmdSetAlarmRev1` shouldn't accept gen5 at all, and that the timestamp should be range-checked before truncation. odd that the new function validates `hapticMode` but not `when`.
> * `app_state.dart:3672` still says the engine computes real sub-seconds "for the rich 20-byte firing form" — that one's missed by the doc sweep.
> 
> **provenance**: there are comments and test names across both prs that say how the bytes were obtained, two of which also name the upstream project, and #32 publishes a band serial. can you strip those to just the vector — "epoch 1781912880 → `01 30 d5 35 6a 00 00 00 00`" says everything useful. pr and issue bodies are editable, worth doing before any of this merges.
> 
> what's right and i want kept: deleting "the rich form is THE form that actually fires" as an unqualified claim, "a GET_ALARM readback proves only that a body was STORED", "56 proves latch, not execution", and removing the fabricated `idx` that was reading an epoch byte as a slot id and inventing `-1` otherwise. gen5 is genuinely untouched in both directions, which i checked.
> 
> (same comment on protocol#33 — they only make sense as a pair.)

You're right about the frames — I rebuilt both forms with the package's own builders and with `hapticMode = 0` they're byte-identical, matching your pair exactly. So short-vs-rev-1 was never a wire distinction and my A/B couldn't have tested it.

That made me re-test the part that IS a wire distinction, cleanly. Today, on my band: laptop-driven, no phone connected, band on wrist, RTC verified at 1 s drift, one arm per trial, latch verified by SET ack + GET_ALARM readback before disconnecting. Verdict signal = post-target readback armed flag (fired ⇒ the band auto-disables, event 59; not fired ⇒ the alarm stays stored and armed — both branches verified on this firmware), plus the buzz on my wrist.

| trial | form        | margin  | latched | executed |
|-------|-------------|---------|---------|----------|
| 1     | rich 20B    | +90 s   | yes     | **no** — still armed past target |
| 2     | rich 20B    | +10 min | yes     | **no** — still armed past target |
| 3     | rev-1/short | +10 min | yes     | **yes** — fired at the armed second, auto-disabled |

The rich payloads carried the exact block your band latched and executed (`47,152,0,0,0,0,0,0 | 0,0 | 7 | 30`), so content is controlled. Raw arms:

```
rich  +90s : 04 00 63 30 87 6a 00 00 2f 98 00 00 00 00 00 00 00 00 07 1e
rich  +10m : 04 00 25 33 87 6a 00 00 2f 98 00 00 00 00 00 00 00 00 07 1e
rev-1 +10m : 01 b3 35 87 6a 00 00 00 00      → executed autonomously
```

So both of our bands are telling the truth: yours executes the rich form, mine latches it and never executes it. Which lands on your hypothesis — firmware version. Mine (`cmdReportVersionInfo`, opcode 0x07, raw reply): fw **41.17.4** (build 0), boot **17.2.2** (build 0), plus `3.5.0.0`, `3.12.1.0` and a trailing `0x608`:

```
01 01 01
29 00 00 00  11 00 00 00  04 00 00 00  00 00 00 00
11 00 00 00  02 00 00 00  02 00 00 00  00 00 00 00
03 00 00 00  05 00 00 00  00 00 00 00  00 00 00 00
03 00 00 00  0c 00 00 00  01 00 00 00  00 00 00 00
08 06 00 00
```

Yours? Two more observables from the run, in case they help your analysis:

- My 56 events carry a compact 8-byte body (`01 00 <epoch u32> 00 10`) with no haptic-block echo even for a rich arm — yours echoed the full block. Another firmware separator.
- The 0x42 RESP and the GET_ALARM readback echo the epoch at a **form-dependent offset** (a rev-1-shaped arm carries it one byte earlier than a rich one — the echo mirrors payload bytes). A fixed-offset parse reads garbage for one of the two forms; same trap that produced the `idx` you flagged.

Given the frames are identical and the firmware split is real, I'd rather settle what the data means before touching either branch — your call on how you want the PRs reshaped from here; happy to follow your lead on that and on your other points.

abdulsaheel:
can you fix conflicts ?

Pablodvs:
@abdulsaheel Rebased and fixed conflicts, are we going to go this way at the end?

Issue #249, opened 2026-08-15
Sleep Tracking (in bed missing) / 5.0


adrnxq:
### Which metric?

Sleep Stages

### What it showed, and why you think it's wrong

When tracking sleep with Whoop through Edge (OpenStrap), the data synced to Apple Health is inaccurate. After comparing the sleep stages generated by Edge with another open-source alternative (⁠noop⁠), it appears the issue stems from Edge missing the "In Bed" (Time in Bed) metric when writing to HealthKit.
Actual Behavior:
Edge exports sleep stages to Apple Health but fails to include the "In Bed" metric. Because Apple Health receives an incomplete data set, third-party dashboards that rely on this data misinterpret the sleep sessions.
For example, when viewing last night's data in Bevel (an external dashboard that fetches data via Apple Health), it incorrectly registered only 3 hours of actual sleep, while categorizing the rest of the night as a "nap."
Expected Behavior:
Edge should include the "In Bed" metric alongside the specific sleep stages when writing to Apple Health. This ensures that the total sleep duration and continuity are accurately reflected and can be correctly parsed by both Apple Health and external apps.
Steps to Reproduce:
1. Sync Whoop sleep data using Edge.
2. Allow Edge to write the sleep data to Apple Health.
3. Open Apple Health (or a third-party app like Bevel) and observe the fragmented/incorrect sleep duration and missing "In Bed" time.
Additional Context:
 The open-source app ⁠noop⁠ handles this sync correctly by consistently passing the "In Bed" metric to HealthKit alongside the other stages. Reviewing how ⁠noop⁠ formats its HealthKit payload might be a helpful reference for a fix.
 Because Apple Health acts as the middleman, the flawed import from Edge cascades into any other health dashboard the user relies on.

### What kind of wrong?

Confident number where the data is missing (should show "—")

### App version

0.9.25+56

### Which band?

WHOOP 5.0 / MG (experimental — expect breakage)

### Screenshots

<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/63278381-de34-46cf-82bc-88c13121a88d" />
<img width="1206" height="1682" alt="Image" src="https://github.com/user-attachments/assets/65c0f956-02b6-42e0-91b0-318320d8ccf6" />
<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/4e07755c-fc98-4dd0-adcf-3c4e00f0368f" />
[...]

Issue #239, opened 2026-08-12
HealtKit not receiving Light/Core sleep data


reziZile:
### What happened?

The data for light sleep stage (core in Apple Health) didn't sync to Apple Health app

### App version

0.9.26

### Platform

iOS (TestFlight)

### Phone model and OS version

iPhone 14 Pro, iOS 26.6

### Which band?

WHOOP 4.0

### Screenshots

<img width="1179" height="1367" alt="Image" src="https://github.com/user-attachments/assets/e4894351-c96e-4286-9b37-b4bccb8d2de5" />
<img width="1179" height="2379" alt="Image" src="https://github.com/user-attachments/assets/8341e69b-fd04-471c-ab04-95ce2f24f43c" />
<img width="1179" height="1605" alt="Image" src="https://github.com/user-attachments/assets/9eceb384-625c-43f1-b087-21900818320b" />

### Quick checks

- [x] The official WHOOP app is not also connected to this band
- [x] I've searched existing issues for this

MD5-Hashm:
I've also had this issue
iPhone 13 Pro, ios 26.6, latest testflight build

abdulsaheel:
Hi this is fixed in latest release, please reopen same issue. if you encounter it again

PR #271, opened 2026-08-20
fix(health): rewrite Apple Health sleep as one in-bed night (Core + stale samples)

### LINKED ISSUE #225: HealthKit Sleep Export Data Truncation and Timestamp Misalignment
### What happened?

OpenStrap Edge writes invalid sleep stage payloads to Apple Health. The core analytics compute the correct sleep duration and stages locally, but the data exported to Apple Health fails to match the local database, resulting in truncated durations and incorrect timestamps.
Expected Result
Apple Health synchronization must identically mirror local app data. A sleep session recorded as 7h 17m (01:06 to 08:42) in OpenStrap Edge must write a contiguous 7h 17m block to Apple Health, preserving all discrete stage transitions (Awake, REM, Light/Core, Deep).
Actual Result
The Apple Health export writes a fragmented, inaccurate record.
 Reference ⁠image_2.png⁠ (Local App Data): Validates a sleep session from 01:06 to 08:42. Total time asleep is calculated as 7h 17m.
 Reference ⁠image.png⁠ (Apple Health Data): Receives only 1h 56m of total sleep time. Timestamps are offset and fragmented, writing REM and Awake stages prior to 11:00 PM (before the actual sleep start time) and recording isolated Deep sleep instances near 01:00 and 03:00.

### Steps to reproduce

1. Record a full sleep session using OpenStrap Edge on iOS.
2. Allow local analytics to process the session.
3. Verify the local sleep duration and stages within the OpenStrap Edge UI.
4. Open Apple Health > Browse > Sleep > Show More Sleep Data.
5. Observe the discrepancy in total duration, stage mapping, and timestamps between the two interfaces.

### App version

0.9.25

### Platform

iOS (TestFlight)

### Phone model and OS version

14 pro max iOS 27 db4

### Which band?

WHOOP 4.0

### Screenshots

Image
Image

### Quick checks

- [x] The official WHOOP app is not also connected to this band
- [x] I've searched existing issues for this

ignaciojuarez:
## Summary

The in-app hypnogram was already right. Apple Health was not: a 7h night showed up as ~2h of REM/Deep, often with no Core, no In Bed, and leftover REM/Awake around 11pm from an earlier onset.

Three things were still true after the plugin bump in #258:

- **Core / In Bed never landed.** `SLEEP_LIGHT` and `SLEEP_IN_BED` go through the Flutter `health` plugin. Android already abandoned that path for a native `SleepSessionRecord` replace. iOS still wrote one sample per stage, so a failed Core/in-bed write left Health Time Asleep as whatever REM+Deep survived (#239, #249).
- **Stale 11pm samples were outside the delete window.** A later derive that moves onset (23:00 → 01:06) leaves the previous HealthKit samples *outside* `[night.start, night.end)`. The day-scoped plugin delete never reached them (#225).
- **Already-exported nights never got another chance.** `health_export_through` skips the finalized prefix. A writer fix does not rewrite last week's Health samples unless that cursor is cleared.

This PR:

- Writes Apple sleep natively, the way Android already does: one `replaceSleepSession` deletes *our* overlapping `sleepAnalysis` samples, then saves one `inBed` envelope plus `asleepCore` / `asleepDeep` / `asleepREM` / `awake`. The plugin is not on this path (unknown keys still map to `bodyMass` and hang `delete()`).
- Uses the same **noon-to-noon** cleanup Android uses, so leftover 11pm fragments die even when onset moved later.
- Fills holes in the in-bed envelope as **awake** (does not invent stages for an unstaged/imported window — those still write only In Bed on Apple).
- **Rewrites older Health data on the next sync.** First export after this update clears `health_export_through` once (`health_sleep_export_epoch = apple-native-1`) so every retained day (up to 400) is written again. Nights that only exist in HealthKit because the day_result is already gone cannot be reconstructed honestly; if those fragments are still a problem I can follow up with a source-scoped delete-only sweep.

Fixes #225, #239, #249.

## Before / after

Same night, 20 Aug 2026, on my phone:

**Edge** — 5h 41m asleep, 12:50 AM–7:17 AM, 6h 28m in bed. Light is most of the night.

OpenStrap Edge sleep

**Apple Health** — 2 hr 36 min Time Asleep. Core row is empty. Stages are fragmented with holes (no awake fill), so Health only counts the REM+Deep slivers.

Apple Health sleep

After this lands: one In Bed bar spanning the detected window, Core for `light`/`nrem`, leftover fragments gone. First Health sync after updating replaces nights already sitting in Health from earlier plugin writes.

## Test plan

- [ ] `flutter analyze` (sleep files clean; repo has one pre-existing info lint in `band_notifications_test.dart`)
- [ ] `flutter test --concurrency=1` — `test/health_sleep_export_test.dart` covers noon-to-noon leftover 11pm, Core/`{t,stage}`/ms timestamps, gap→awake, epoch cursor reset, native Apple channel payload including empty-stage In Bed
- [ ] On a device that already has the truncated nights: update, grant Health if needed, foreground sync, then Apple Health → Browse → Sleep. Last night should match the Edge hypnogram (duration, Core, In Bed). Older retained nights should rewrite on that same first sync.
- [ ] Confirm a night with no hypnogram (CSV import) still does not fail the rest of the day's export, and on iOS writes In Bed only.

### DIFF
--- a/ios/Runner.xcodeproj/project.pbxproj
+++ b/ios/Runner.xcodeproj/project.pbxproj
@@ -21,6 +21,7 @@
 		53962E962FF6EE120061A61B /* WatchBridge.swift in Sources */ = {isa = PBXBuildFile; fileRef = 53962E952FF6EE120061A61B /* WatchBridge.swift */; };
 		53962E972FF6EE120061A61B /* OpenStrapIntents.swift in Sources */ = {isa = PBXBuildFile; fileRef = 53962E942FF6EE120061A61B /* OpenStrapIntents.swift */; };
 		60DE9D949573401269D6DF2E /* HealthRoutes.swift in Sources */ = {isa = PBXBuildFile; fileRef = 46A2B400A52A2CA90242C195 /* HealthRoutes.swift */; };
+		B9D2F406183A5C7E92B1D3F5 /* HealthKitSleepWriter.swift in Sources */ = {isa = PBXBuildFile; fileRef = A8C1E3F507294B6D81A0C2E4 /* HealthKitSleepWriter.swift */; };
 		630172CC9317145AD5F8F3B7 /* Pods_RunnerTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 2C721EA0C8D31A3834E66203 /* Pods_RunnerTests.framework */; };
 		74858FAF1ED2DC5600515810 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 74858FAE1ED2DC5600515810 /* AppDelegate.swift */; };
 		7884E8682EC3CC0700C636F2 /* SceneDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7884E8672EC3CC0400C636F2 /* SceneDelegate.swift */; };
@@ -102,6 +103,7 @@
 		331C8081294A63A400263BE5 /* RunnerTests.xctest */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = RunnerTests.xctest; sourceTree = BUILT_PRODUCTS_DIR; };
 		3B3967151E833CAA004F5970 /* AppFrameworkInfo.plist */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.plist.xml; name = AppFrameworkInfo.plist; path = Flutter/AppFrameworkInfo.plist; sourceTree = "<group>"; };
 		46A2B400A52A2CA90242C195 /* HealthRoutes.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; path = HealthRoutes.swift; sourceTree = "<group>"; };
+		A8C1E3F507294B6D81A0C2E4 /* HealthKitSleepWriter.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; path = HealthKitSleepWriter.swift; sourceTree = "<group>"; };
 		5348973B2FDC19C80033A4D9 /* OpenStrapWidgetExtension.appex */ = {isa = PBXFileReference; explicitFileType = "wrapper.app-extension"; includeInIndex = 0; path = OpenStrapWidgetExtension.appex; sourceTree = BUILT_PRODUCTS_DIR; };
 		5348973C2FDC19C80033A4D9 /* WidgetKit.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = WidgetKit.framework; path = System/Library/Frameworks/WidgetKit.framework; sourceTree = SDKROOT; };
 		5348973E2FDC19C80033A4D9 /* SwiftUI.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = SwiftUI.framework; path = System/Library/Frameworks/SwiftUI.framework; sourceTree = SDKROOT; };
@@ -299,6 +301,7 @@
 				ACCE55E7000000000000F11E /* AccessorySetup.swift */,
 				0BGTASK00000000000000002 /* BgSyncScheduler.swift */,
 				46A2B400A52A2CA90242C195 /* HealthRoutes.swift */,
+				A8C1E3F507294B6D81A0C2E4 /* HealthKitSleepWriter.swift */,
 			);
 			path = Runner;
 			sourceTree = "<group>";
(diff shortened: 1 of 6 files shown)


ignaciojuarez:
Same night on device (20 Aug 2026). Edge has the full hypnogram; Health does not.

**Edge** — 5h 41m asleep, 12:50 AM–7:17 AM, 6h 28m in bed. Light (Core) is present.

OpenStrap Edge sleep

**Apple Health** — 2 hr 36 min Time Asleep. Core is empty. Only REM/Deep slivers survived.

Apple Health sleep

abdulsaheel:
the #225 half of this is real and i had it wrong — i assumed #258 fixed it and it didn't. main's window starts at `dayStart` when onset moves later, so 23:00 leftovers are never reached, and the plugin delete used `.strictStartDate` so it wouldn't have caught them from an overlapping window either. noon-to-noon plus overlap semantics fixes both. that part i want.

but it doesn't build, and there are two things i can't take as written.

**1. the ios app doesn't compile.** `HealthKitSleepWriter.swift:126-128` uses `.asleepREM` / `.asleepCore` / `.asleepDeep`, all ios 16+, and the runner target is `IPHONEOS_DEPLOYMENT_TARGET = 15.0` (`project.pbxproj:774`, and `platform :ios, '15.0'` in the podfile).

```
'asleepREM' is only available in iOS 16.0 or newer   :126:24
'asleepCore' is only available in iOS 16.0 or newer  :127:26
'asleepDeep' is only available in iOS 16.0 or newer  :128:25
```

`flutter build ios --debug --no-codesign` succeeds on main and fails on this head. fix is an availability guard falling back to `.asleep` on 15 (that's what pre-16 healthkit has), or raise the target — your call, but a guard is the smaller change. worth knowing analyze is clean and all 2833 tests pass on this branch: nothing in dart-land can see a swift compile error, which is why i asked for a real xcodebuild.

**2. `unobserved` is exported as `awake`, and that fabricates a measurement.** `health_sleep_session.dart:193` plus the gap-fill in `normalizeHealthSleepSession`. `onehz_pipeline.dart:1720` says it outright: "unobserved is not a stage and not wake — it is a second the band did not watch, and it belongs to neither side". seconds we didn't watch get published to apple health and health connect as measured wake, and other apps read that as fact.

and it doesn't buy the thing it's justified by — apple's time asleep is `asleepCore + asleepDeep + asleepREM`, so awake adds zero to it. the 2h36m → full night comes entirely from writing core at all. drop the `'unobserved'` case and the `append(..., awake)` fill and you lose nothing the pr claims.

that's also an android regression, since `normalizeHealthSleepSession` is shared and `HealthConnectSleepWriter.buildStage` maps awake → `STAGE_TYPE_AWAKE`. your own test change shows it — the android assertion moved `awake: 32 → 42` because a 10 min unlabelled hole is now filled as wake.

**3. the cleanup window can delete a previous night, and it's not recoverable.** `health_sleep_session.dart:242-252` unions noon-to-noon with the whole calendar day. android only does noon-to-noon. for a night that wakes after local noon:

- d−1 night 02:00→13:30, samples at d−1 02:00–13:30
- d night 01:00→09:00 → cleanup `[d−1 12:00, d 12:00)` ∪ `[d 00:00, d+1 00:00)` = `[d−1 12:00, d+1 00:00)`
- the swift predicate is `predicateForSamples(withStart:end:options: [])` — no `.strictStartDate`, so it matches on overlap and deletes the **whole** d−1 envelope sample, not a clipped part
- d−1 is behind `health_export_through`, so it never gets rewritten

now that native owns the delete, just use `sleepSessionCleanupRange` alone like android does and drop the day union.

**4. dst.** `health_sleep_session.dart:236` — `add(const Duration(days: 1))` on a 25 hour fall-back day doesn't advance the date, so `calculatedStart` lands a day early and widens the window into the previous night (which feeds 3). `health_export.dart:729` has a comment specifically about this trap, and the kotlin you ported from uses `plusDays(1)`. one token: `DateTime(localEnd.year, localEnd.month, localEnd.day + 1)`.

**5. the epoch reset isn't gated on apple.** `health_export.dart:103-113`, called at `:488` — it clears `health_export_through` on every platform, so android replays up to 400 day bundles on the first sync after update, re-running type deletes, hourly energy writes and the minute-hr batch. nothing in the health connect writer changed in this pr. gate it on `isApple`. everything else about the cursor checks out — idempotent, can't loop, can't duplicate, and a partial failure leaves it sane.

on the premise: bullet 1 of your description says core and in-bed still don't land after the plugin bump. #258 merged 7h before this opened and the screenshots are of the night of 19→20 aug, so that night was slept on a build without it. i went through `health 12.2.1` and couldn't find a mechanism — `_alignValue` maps `SLEEP_LIGHT → 3` and the swift writer builds the category sample straight off that, which is `.asleepCore`. can you reproduce "no core" on a build that actually contains #258? bullets 2 and 3 stand either way.

things i checked and they're fine, so you don't have to defend them: the delete predicate is correctly scoped — `HKSource.default()` is this bundle only, not the watch, not manual health entries, not another tracker, and healthkit refuses to delete objects an app didn't save anyway. the pbxproj is 4 clean additions with no version strings touched. no derived number moves. and removing every `SLEEP_*` from `healthDeleteTypes` closes a real hazard — the plugin's `delete` never calls `result()` on an unsupported type, so that await used to hang rather than fail. every path in your swift `replace()` calls completion exactly once.

last thing, minor: most of the 356 changed lines in the test file are `dart format` churn, and it hides the three real assertion changes. we don't run it in this repo — the old style and the sdk's tall-style formatter reflow thousands of lines against each other. worth reverting the reformat so the diff is just what changed.

PR #365, opened 2026-09-08
fix: hold the beat axis non-decreasing (sleep staging threw on 6 of 8 days), bump kAlgoVersion to 88; quiesce ingest during a data reset

DropTabl:
## What broke

`beat_ts_ms` is a **modelled** beat position. `beatTimesMs` anchors on the record's sub-second stamp — under the model that the record's *last* beat sits at that stamp — then walks the intervals backwards from it. When two adjacent records' anchors sit closer together than one beat, the model places record N+1's beat 0 **earlier** than record N's last beat. Beats are emitted in record order, so the axis steps backwards.

Measured on a live install, one day's derive window:

| beats | inversions | where | by |
|---|---|---|---|
| 39,066 | 94 | every one at `beat_index = 0` | 7–672 ms |

`cardioStager` binary-searches that axis and asserts it ascends. One inverted pair takes the whole night down:

```
derive day 2026-09-08 FAILED/skipped: Bad state: sleep-staging isolate failed:
'rrTsMs must be non-decreasing — the beat window is binary-searched'
```

Six of eight days in the pending span threw on **every** pass. The newest day that survived was the previous night, so the app sat on *"Crunching last night's numbers"* and tapping Sync again looked like the only lever. The BLE offload was healthy the whole time — `OFFLOAD SUMMARY: complete=true`, reaching the live edge every pass. The sync was never the problem.

## Why `kAlgoVersion` moves

The assert is **compiled out in release**. Shipped builds never threw — they mis-selected each 300 s window's beats and banked the result, which is the worse of the two failures. Those days cannot be told apart from good ones after the fact, and `finalizedDayIds` / `sleepSessionCandidate` both key on the version, so without a bump they are served from cache forever. 87 → 88; `kAnalyticsPin`/`kProtocolPin` unchanged, the repair is entirely edge-side.

## Why a clamp and not a sort

**The order is not what is wrong.** Beats arrive in the order the strap detected them, and that order *is* the rhythm — `beat_clock_read_path_test` pins it as the contract it is. The first attempt at this sorted the `(rrTsMs, rrMs)` pair and that test correctly killed it.

So `monotonizeBeatAxis` repairs only the modelled *placement*, holding each beat no earlier than the one before it. Interval values and their order are untouched. A held beat lands on a zero-length gap, which the axis already contains — the `rr_ts_ms` staircase ties a whole record's beats to one millisecond — and which analytics' window gather is explicitly written for (its lower bound is inclusive so tied beats are not dropped). Verified against the consumers: RMSSD is computed from interval *values*, and the Lomb–Scargle denominators are guarded.

Called on both paths into `Substrate.rrTsMs`: at the end of `decodeSubstrate`, and at the page seam in `buildSubstrate()` — a seam is invisible to either page loader, since each places its own beats correctly and then appends.

## Also: "Synced through 11:06" on Home

The data edge was only readable from Profile › Devices, so there was no way to see how far a sync had actually got. It reads the band's own clock on the newest record **banked**, not last contact — a connection that transfers nothing is not progress and must not render as progress.

"Today" is the day the screen is **already showing** rather than `DateTime.now()`, so it cannot contradict the date line directly above it after midnight, and needs no timer. It renders on the loading / failed / first-run path too, which is when the question is loudest.

Two fixes fall out of putting that value on screen, both pre-existing:

- `_onRecord` advanced `_lastRecTs` *before* awaiting the insert. The field only moves forward, so a failed write advertised a record the database does not hold — for the rest of the process.
- `resetAllData()` wiped the database but left `_lastRecTs` set, so a re-pair without a relaunch showed the deleted install's timestamp.

## Tests

`test/beat_axis_consumer_contract_test.dart` drives the **real** `cardioStager` over a 120-record fixture whose anchor drift genuinely wraps — with a guard test asserting the fixture actually inverts, so it cannot quietly stop testing anything. Plus unit coverage for the clamp and the Home formatter.

Verified the tests fail without the repair: **3 failures, including the real assertion**. Full suite on this branch: **3673 passed, 446 skipped, 0 failed**.

## Summary by Sourcery

Repair beat-axis ordering, recompute affected analytics with algorithm version 88, expose the banked sync edge on Home, and quiesce ingest while deleting data.

New Features:
- Display the banked data edge on Home with a contextual “Synced through …” status for loading, failed, first-run, and normal states.

Bug Fixes:
- Keep modelled beat timestamps non-decreasing without changing beat order or interval values, preventing analytics window-selection failures and silent release mis-selection.
- Prevent record-ingest and acknowledgements from repopulating the database during data resets, including foreground and headless sync paths.
- Ensure the Home data-edge status advances only after successful persistence and is cleared during reset.

Enhancements:
- Bump the algorithm version from 87 to 88 so affected derived days are recomputed instead of served from cache.
- Add a process-wide, nested reset gate covering all database write paths and preserving safe band retry behavior when commits are refused.

Tests:
- Add consumer-level and unit coverage for beat-axis monotonicity, preserving interval order, Home sync formatting, and reset-ingest quiescence.

## Summary by CodeRabbit

- **New Features**
  - Added a Home screen status showing how far band data has synced, including support for days with no received data.

- **Bug Fixes**
  - Corrected beat timing order to improve analytics window selection.
  - Recalculated affected days to reflect corrected results.
  - Resetting app data now clears outdated sync information and temporarily pauses incoming records, including background sync.
  - Sync status no longer advances when saving a record fails.

### DIFF
--- a/lib/compute/derivation_engine.dart
+++ b/lib/compute/derivation_engine.dart
@@ -1612,7 +1612,25 @@ import 'substrate.dart';
 // had) in exactly the shape that kept "readiness —" live for three releases.
 // kAnalyticsPin/kProtocolPin are UNCHANGED: M5 touches no analytics or
 // protocol code.
-const int kAlgoVersion = 87;
+// v88 — THE BEAT AXIS IS HELD NON-DECREASING (`monotonizeBeatAxis`,
+// substrate.dart). `beat_ts_ms` is a modelled beat position walked backwards
+// from the record's sub-second anchor, so two records whose anchors sit closer
+// than one beat place N+1's first beat before N's last. Measured on a live
+// install: 94 inversions across 39,066 beats in a single day.
+//
+// Analytics binary-searches that axis and asserts it ascends. The assert is
+// compiled out in RELEASE — so shipped builds did not throw, they mis-selected
+// each 300 s window's beats, and every affected day was banked at v87 from a
+// window gather that had silently skipped or duplicated beats. Those results
+// cannot be told apart from good ones after the fact, which is exactly what a
+// version bump is for: the days re-derive instead of being served from cache
+// (`finalizedDayIds` / `sleepSessionCandidate` both key on kAlgoVersion).
+//
+// Real output change on affected days — window membership moves by up to the
+// inversion depth (672 ms measured) — so the bump is real. Interval VALUES and
+// their order are untouched by the repair; only the modelled clock moves.
+// kAnalyticsPin/kProtocolPin are UNCHANGED: the repair is entirely edge-side.
+const int kAlgoVersion = 88;
 /// The sibling SHAs this version was derived against, asserted against
 /// pubspec.yaml in test/db_serve_version_and_reads_test.dart.
 ///
--- a/lib/compute/derive_prepare.dart
+++ b/lib/compute/derive_prepare.dart
@@ -689,6 +689,11 @@ class _PrepareAccumulator {
   }
 
   Substrate buildSubstrate() {
+    // The page seam, which neither page loader can see: `addDecodedPage` and
+    // `addRawPage` each place their own beats correctly and then APPEND, so an
+    // overlap between the last record of one page and the first of the next is
+    // only visible once every page is in. See [monotonizeBeatAxis].
+    monotonizeBeatAxis(rrTsMs);
     // `_skinTempUnitDrops` (see [_skinTempFor]) is structurally 0 today —
     // `derivableSourceSql()` renders `source IS NULL` (db.dart:116), so a
     // day's admitted rows are all one family and never trip the guard.
--- a/lib/compute/substrate.dart
+++ b/lib/compute/substrate.dart
@@ -777,6 +777,10 @@ Substrate decodeSubstrate(List<String> hexes) {
       ..clear()
       ..addAll(sr);
   }
+  // The beats this path just placed can still overlap across a RECORD seam —
+  // see [monotonizeBeatAxis]. The sort above only runs when loose live beats
+  // were folded in, so it is not that guarantee.
+  monotonizeBeatAxis(rrTsMs);
 
   return Substrate(
     tsSec: tsSec,
@@ -799,6 +803,39 @@ Substrate decodeSubstrate(List<String> hexes) {
     // `hrValidAt` regardless.
     hrValid: List<int>.filled(n, -1),
   );
+}
+
+/// Hold a beat time axis non-decreasing — WITHOUT moving a single interval.
+///
+/// `beat_ts_ms` is a RECONSTRUCTED beat position: `beatTimesMs` anchors on the
+/// record's sub-second stamp under the model that the record's LAST beat sits
+/// at that stamp, then walks the intervals backwards from it. The anchor is
+/// measured; the placement is modelled. When two adjacent records' anchors sit
+/// closer together than one beat, the model puts record N+1's beat 0 EARLIER
+/// than record N's last beat, and beats are emitted in record order — so the
+/// axis steps backwards. Real, not theoretical: 94 inversions across 39,066
+/// beats in one day on a live install, every one at `beat_index = 0`, by
+/// 7-672 ms.
+///
+/// Analytics binary-searches this axis (`_cleanBeatsInWindow`) and asserts it
+/// is non-decreasing. That assert is compiled out in release, so a release
+/// build does not throw — it silently mis-selects each window's beats instead,
+/// which is the worse failure of the two.
+///
+/// THE ORDER IS NOT WHAT IS WRONG. Beats arrive in the order the strap detected
+/// them and that order IS the rhythm — `beat_clock_read_path_test` pins it as
+/// the contract it is. Only the modelled placement is repaired, and by the
+/// least it can be: each beat is held no earlier than the one before it. A sort
+/// would reshuffle a true interval series to satisfy an assert.
+///
+/// A held beat lands on a zero-length gap, which the axis already allows and
+/// already contains: the `rr_ts_ms` staircase fallback puts a whole record's
+/// beats on one millisecond, and analytics' window gather is written for those
+/// ties (its lower bound is inclusive precisely so tied beats are not dropped).
+void monotonizeBeatAxis(List<double> rrTsMs) {
+  for (var i = 1; i < rrTsMs.length; i++) {
+    if (rrTsMs[i] < rrTsMs[i - 1]) rrTsMs[i] = rrTsMs[i - 1];
+  }
 }
 
 /// Steps MEASURED by the band's own pedometer over [sub], or `null` when this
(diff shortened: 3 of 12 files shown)


DropTabl:
Both review findings addressed in `dd124aa`. Thanks — the second one was a real bug that defeated part of this PR's own fix.

### CodeRabbit: quiesce record delivery before the wipe

Correct, and it undid what this PR added. `resetAllData()` clears `_lastRecTs`, but its documented order puts `signOut` — and the `unpair` inside it — **last**, deliberately, because that flips the route and unwinds the UI. The strap therefore stays connected and keeps handing over historical records straight across `wipeAll()`. One landing after it is written into the fresh database and re-advertised as the data edge: the deleted installation, visible again on Home.

Tracing it turned up a second path the suggestion didn't name, and it's the one that carries the volume:

```dart
onRecord:       _onRecord                       // singles
onRecordsBatch: LocalDb.insertRecordsBatch      // the drain — straight to the db
```

The drain is wired directly to `LocalDb`, so it bypasses every check AppState makes. Guarding only `_onRecord` would have looked like a fix and left the real hole open. Both now refuse while `_resetting` is up.

I used a refusal flag rather than reordering the teardown: tearing the link down first is a change to the **connection lifecycle**, not to the wipe, and that ordering is deliberate and load-bearing (the docstring says so). Nothing is lost by refusing — none of these records have been ACKed, so they are still on the band and arrive on the next sync. The flag is raised before the wipe and lowered in a `finally`, because a half-reset install that silently drops every record would be worse than the race it closes.

### Sourcery: `insertRecord`'s return value

Applied. One correction for the record: the trigger described — "fails non-throwingly and returns `false`" — is **not reachable today**. `insertRecord` sets `inserted = true` as the last statement inside the transaction and never catches, so it can only return `true` or throw, and the ordering fix already handled the throw. But this PR's whole claim is that the line must never advertise data we do not hold, and ignoring the one value that reports whether we hold it made that promise on trust rather than on evidence. Cheap to make honest.

### Test

`test/reset_quiesces_ingest_test.dart` greps rather than runs: `_resetting` is private and `resetAllData` needs the entire plugin stack (preferences, notifications, widget, telemetry, keychain), while the failure mode is *a third ingest path added later that forgets the guard* — invisible to any test that exercises the code. Same approach and same reasoning as `link_priority_structural_test.dart`. It pins both guards, the raise-before-wipe ordering, the `finally`, and that `insertRecordsBatch` is never handed back as a bare tear-off.

Full suite: **3677 passed, 446 skipped, 0 failed.**

DropTabl:
Pushed `69e45dd` — the second half of the reset quiesce. This closes the same hole in the headless engine, and I'd rather it land here than as a follow-up: `dd124aa` fixed the foreground half, and shipping only that leaves the identical defect in the other engine. Half a quiesce is a worse thing to merge than none.

### What was still open

`runHeadlessSync` (background_sync.dart) builds a **second** `BleEngine` whose sinks go straight to `LocalDb`, with no `AppState` anywhere in scope — so nothing `AppState` guards can reach it. A drain already in flight when the user confirms "delete everything" kept writing into the freshly wiped database.

It's narrow: a foreground engine preempts a background drainer when it connects (`BleEngine._claimBand`) and `AppState` connects on launch, so you need a drain the foreground never preempted. But the only thing standing between a reset and a *fresh* headless drain today is an accident — `prefs.clear()` makes `PairedDevice.load()` return null, so the next run bails. Nobody wrote that as a safeguard, and it stops working silently the day the paired device is cached anywhere else.

### Why a static is legitimate here

`ResetGate` replaces the private `_resetting` field so both engines read one flag. That only works because there is no second isolate to hide from it — every headless entry point runs in the **main** isolate:

- Android post-boot: `EdgeApplication.onCreate` pre-warms the FlutterEngine and runs `main()` with no Activity
- iOS CoreBluetooth restore wake, and both BGTask profiles

and the WorkManager tasks that *did* have their own isolate were pulled for exactly that class of problem — `main()` still cancels them by unique name on every Android start, and nothing registers a `callbackDispatcher`. `HeadlessSyncGate` already rests on the same assumption. `reset_gate.dart` states this explicitly, including that reintroducing a background isolate breaks the mechanism **silently**.

### The part worth reviewing closely

**The commit sink throws; it does not return quietly** — in both engines.

Only `onCommit` banks raws + archives + trim cursor in one transaction, and `DrainController` reads durability *from a throw*. Its safe-trim invariant even forbids `onRecordsBatch` without `onCommit`, to close "the latent hole where archive-only chunks would ACK and trim flash that was never stored".

So a silent `return` in the commit gate would tell the drain the chunk was banked → it ACKs → **the band trims flash this reset refused to store**. That converts a race into real data loss, on a band whose copy is the only surviving one if the reset then fails. A throw blocks the ACK and the records stay on the strap.

The non-ACK sinks (`onRecord` / `onRecordsBatch` / `onEvent` / `onArchiveRecord`) return quietly, which is safe precisely because none of them gate the trim.

### Tests

`test/reset_quiesces_ingest_test.dart` grows to cover the second engine: every write sink consults the gate, no sink is a bare `LocalDb` tear-off again, the headless bail sits ahead of `PairedDevice.load()`, and both commit sinks throw.

Full suite at CI's invocation (`flutter test --concurrency=1`): **3682 passed, 446 skipped, 0 failed.**

Note for anyone running the suite in parallel: it races on real DB files and produces unrelated flakes (`ui2_wellness_recovery_paint_test`, `oura_link_test` for me). Both pass 3/3 in isolation on this branch and on its base — `--concurrency=1` is the one that means anything.

DropTabl:
`7464c75` — applied. `ResetGate` now counts resets instead of latching a bool.

Verified the trigger is reachable rather than theoretical: `resetAllData()` is awaited but does **not** block the UI. The confirm dialog is already dismissed when it starts (`settings.dart` pops it, *then* calls), `backToRoot(c)` only runs when it returns, and wiping a real database takes seconds. So a user can walk back into Settings and confirm a second wipe on top of the first — and the first to finish would lower the flag while the second is still wiping, reopening the exact window this class exists to close, at the worst possible moment.

`leave()` clamps at zero rather than going negative. An unbalanced `leave` is a bug either way, but one that drove the count below zero would make the **next** reset's `enter()` fail to open the gate at all — and that is the direction that loses data. Absorbed rather than asserted.

Three tests added: nesting, and that an unbalanced `leave` cannot drive the gate below shut.

---

One note on the earlier finding, for the record. The 12:24 review recommended the `onCommitBatch` gate *"return immediately"*. I implemented a **throw** instead, deliberately — `CommitSyncBatchSink` is a `Future<void>` sink and `DrainController` reads durability *from a throw*; its own safe-trim invariant exists to close "the latent hole where archive-only chunks would ACK and trim flash that was never stored". A quiet return there would read as a successful commit, the drain would ACK, and the band would trim flash the reset refused to store — converting the race into real data loss, on the only surviving copy if the reset then fails. The non-ACK sinks do return quietly, which is safe precisely because none of them gate the trim.

Full suite at CI's invocation (`flutter test --concurrency=1`): **3684 passed, 446 skipped, 0 failed.**

PR #399, opened 2026-09-18
gate the engine state callback behind resetgate too

abdulsaheel:
same bug as #388 (_onLiveEvent) but for _onEngineState, the onState callback wired into BleEngine. it fires ~1hz during any live connection and writes straight to LocalDb (device/battery rows) plus PairedDevice's prefs namespace — no _resetting check anywhere in it.

resetAllData keeps the band connected on purpose while it awaits wipeAll/prefs.clear (comment right there says so). since this one was ungated, a tick landing in that window resurrects a device/battery row or rewrites the pairing record right after it was just wiped/cleared. "delete everything" ends up not actually empty, or worse, silently re-pairs.

fix is the same one-line guard as the other callbacks. added a grep-pinning test alongside the existing ones in reset_quiesces_ingest_test.dart so a future new write path in this handler can't regress it quietly.

## Summary by Sourcery

Quiesce engine-state ingestion during data resets to ensure the reset leaves local data and pairing preferences cleared.

Bug Fixes:
- Prevent engine-state updates from repopulating device data or pairing preferences while resetAllData is in progress.

Tests:
- Add regression coverage ensuring the engine-state callback is gated during resets.


abdulsaheel:
duplicate of #397 (already merged) — the finder that surfaced this ran against main before #397 landed. closing, not a real gap.

